PR #7, opened 2019-12-04
Document the EOF behaviour for synchronous file handles with lpOverlapped

macdice:
The documentation doesn't seem to cover the EOF behaviour when a synchronous file handle is accessed with an OVERLAPPED struct.  That is a useful thing to do, in order to provide an offset, to simulate the POSIX function pread() (ie to avoid having a separate lseek() syscall).  See also Raymond Chen's blog[1].

[1] https://devblogs.microsoft.com/oldnewthing/20150121-00/?p=44863


PRMerger12:
@macdice : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:
Thanks for your contribution @macdice.  I've merged your changes and they should be live soon.

PR #10, opened 2019-12-12
NETRESOURCE URL fix

mwilliams31:


GrantMeStrength:
I will fix - but remove en-us

GrantMeStrength:
Thanks! I will correct. Need to remove en-us.

PR #15, opened 2019-12-19
Update nf-sysinfoapi-getintegrateddisplaysize.md

zeusk:
Corrected info, the function is located in onecore.lib

Without onecore.lib, calling this function results in undefined symbol. Also, defining _WIN32_WINNT is not necessary to make use of this function.


Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #17, opened 2019-12-20
Fix bad escape

nuno-andre:


PRMerger13:
@nuno-andre : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #19, opened 2019-12-22
â€“ seems like something like "i.e." for clarify

josephshen:


KalleOlaviNiemitalo:
â€“ is what you get if you encode U+2013 EN DASH ("–") as UTF-8 and then decode it as Windows-1252.

stevewhims:
Thank you, @josephsmeng. :)

PR #23, opened 2019-12-29
Corrected description of FirstElement in the D3D11_BUFFER_SRV structure

jjhegedus:
Corrected description of FirstElement in the D3D11_BUFFER_SRV structure.  It is not a count of bytes before the first element but rather the index of the first element using the stride.


stevewhims (approved):
Thank you!

PR #27, opened 2020-01-02
Escape backslashes in GetDiskFreeSpace[Ex] docs

MSDN-WhiteKnight:
Escape backslash characters because otherwise they are not rendered (or single backslash is rendered in place of double).


GrantMeStrength:
thank you!

PR #38, opened 2020-01-15
Escape backslashes in extended length prefix

benmcmorran:
This was rendering as \\? instead of \\\\?\\


GrantMeStrength:
thanks!

PR #46, opened 2020-01-20
Replaced visible `<i>` with a markdown code syntax

jeremyVignelles:
I didn't understand the doc at first, because the \ was missing and the `<i>` was visible. This change clarifies the syntax.
If you prefer to keep the italics semantic instead, I'm fine with that too, but fix that properly.


PRMerger12:
@jeremyVignelles : Thanks for your contribution! The author(s) have been notified to review your proposed change.

namazso:
you could've simply escaped the backslashes by inserting another backslash, which in turn would make the italics work

jeremyVignelles:
Feel free to open another PR and close this one then. I don't have any strong feeling on that matter.

hkhmsft:

PRMerger9:
@lastnameholiu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #48, opened 2020-01-22
Clarify behavior of GetWindowTextLength towards zero-length titles

sylveon:
I've recently hit that undocumented case myself, so I thought it would be great to add them to the docs to make the behavior of the function less confusing.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-getwindowtextlengtha.md
+++ b/sdk-api-src/content/winuser/nf-winuser-getwindowtextlengtha.md
@@ -73,23 +73,18 @@ A handle to the window or control.
 
 ## -returns
 
+Type: <b>int</b>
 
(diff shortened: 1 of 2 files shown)


PRMerger9:
@sylveon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
thanks @sylveon

Karl-Bridge-Microsoft:
#sign-off

sylveon:
> This function does not clear the most recent error information. To determine success or failure, clear the most recent error information by calling <a href="https://docs.microsoft.com/windows/desktop/api/errhandlingapi/nf-errhandlingapi-setlasterror">SetLastError</a> with 0, then call <a href="https://docs.microsoft.com/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a>.

This might be a bit confusing, I think that it should be pointed out that SetLastError must be called before GetWindowTextLength and that GetLastError should called after.

PR #56, opened 2020-01-27
Missing backslash in volume GUID syntax

AssortedBits:
See standard Volume GUID syntax described elsewhere.  The function fails without the second slash.
See also companion PR: #57


PRMerger14:
@AssortedBits : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #79, opened 2020-02-09
Deep link to Format Conversion using Direct3D 10.1 section

AdamJMiles:
This link needs to jump to the right section of the linked page, not the top.


PRMerger18:
@AdamJMiles : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

stevewhims:
Looks good, thank you! :)

Btw, there are many cases in the current docs that don't ahere to them, but here are some guidelines to be aware of:

- Omit language tags (for example, en-us, or en-gb) from urls.
- If possible, use site-relative links (for example "/windows/win32..." rather than "https://docs.microsoft.com/windows/win32...").
- If possible, use markdown in markdown files. The only place I know of where you can't use markdown is in the description of an enum constant within a Win32 file. Also, `&mdash;` is the only way I know to insert an emdash in markdown. Apart from that, markdown should work.
- Use sentence case in titles (rather than title case)
- In general, prefer informal contractions (for example "doesn't" rather than "does not").

PR #82, opened 2020-02-10
Update nf-winsvc-createservicew.md

xiaoyinl:
See #46. I think the convention is to only italicize variable names. The `ServiceName` should be changed to the actual name of the service, but `NT SERVICE` should not, so it seems we'd better not italicize `NT SERVICE`.


PRMerger13:
@xiaoyinl : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:
Thanks for your contribution @xiaoyinl.  I've merged your changes and they should be live soon.

PR #93, opened 2020-02-14
fileapi: Escape backslashes for \\?\

jay:
~~~
Prior to this change the MS documentation for certain API functions
showed incorrect "\?" for the extended-length path prefix instead of
the correct "\\?\", because the backslashes were not escaped.

**This is only a partial fix**, this is as much as I can do. There are
probably many more functions that need to have \\?\ escaped in certain
areas. Also \\.\ probably needs to be escaped as well.

Closes #xxxx
~~~
---

For example from [CreateFileA](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea):
Capture

All the changes are pretty much to fix that except for the `\\?\Volume...` changes in nf-fileapi-flushfilebuffers.md and nf-fileapi-getvolumepathnamesforvolumenamew.md which I'm not really sure about.


PRMerger20:
@jay : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@lastnameholiu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks, @jay for pointing this out. I went ahead and merged your changes. The scope of these changes is rather big and we're working to continue fixing these typos. Thanks once again.

PR #96, opened 2020-02-17
fixed errors caused by failure to escape

AssortedBits:
fix errors in template format string caused by failure to escape backslashes


PRMerger15:
@AssortedBits : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #102, opened 2020-02-24
lint: Escape backslash separators

Chris-Johnston:
This change escapes the backslash separator between `ServerName\ShareName` and `DomainName\UserName`. Because this was not escaped, it was causing one of the `<i>` tags to show, instead of the slash itself.

TODO:
- [ ] ~~Escape the same backslashes in the CredUIPromptForCredentialsA doc~~ Will do in another PR


PRMerger8:
@Chris-Johnston : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

Chris-Johnston:
Thanks @lastnameholiu ! I was going to go back and fix this same issue in the CredUIPromptForCredentialsA doc too, I guess I'll just do it in another PR.

PR #122, opened 2020-03-09
Update line breaks inside table

phuclv90:
Hello,

I've just seen many broken lines in the table, therefore some texts became a code block due to markdown. That makes it hard so I've fixed it. I have yet no idea why texts in the table are not wrapped properly are broken incorrectly, making the text become a code block. I've also fix some space errors


PRMerger6:
@phuclv90 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #128, opened 2020-03-12
Mention MKSYS_URLMONIKER in MKSYS enumeration?

phimuemue:
MKSYS_URLMONIKER can be obtained from IsSystemMoniker (see https://docs.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-imoniker-issystemmoniker), but is not mentioned in the linked MKSYS enumeration documentation.

I see that it is not defined in the enum (at least not in my ObjIdl.h), but it is defined in urlmon.h and seems to fill exactly the gap in ObjIdl.h.

### DIFF
--- a/sdk-api-src/content/objidl/ne-objidl-mksys.md
+++ b/sdk-api-src/content/objidl/ne-objidl-mksys.md
@@ -106,6 +106,9 @@ Indicates the system's terminal server session moniker class.
 
 Indicates the system's elevation moniker class.
 
+## Remark
+
+MKSYS_URLMONIKER is an additional value.
 
 ## -see-also
 


PRMerger16:
@phimuemue : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for your submission. However, MKSYS_URLMONIKER is not part of the MKSYS enumeration (as defined in objidl.h; you can see it is commented out of the enum definition in objidl.idl, which also ships in the Windows SDK). Therefore, it does not belong in the article you updates, so I'm going to close this pull request.

Instead, MKSYS_URLMONIKER is a standalone #define value in urlmon.h and urlmon.idl (that is, it's not defined in the context of an enum). Given this, I'll update the [IsSystemMoniker](https://docs.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-imoniker-issystemmoniker) article to indicate where this value is defined vs. the other MKSYS values referenced in that article.

PR #133, opened 2020-03-13
Delete garbage file temp.txt from the root of the source tree

bgianfo:
Looks like this was introduced erroneously as part of 22b7ad54b0cff6ad6c95f5590d8560ecac7d3a04


bgianfo:
@GrantMeStrength looks like you accidentally committed this file, can I get a review?

bgianfo:
@drewbatgit can you take a look at this one too?

PR #134, opened 2020-03-13
Delete zero byte file diff-11-20.txt from root

bgianfo:
Looks like this was introduced in 2018 during initial import?
Doesn't appear useful, and should be deleted.


bgianfo:
@drewbatgit Looks like you accidentally committed this file? Can I get a sign off?

PR #163, opened 2020-03-31
Update nf-winsock2-wsapoll.md

compnerd:
- Fix the list element tags for elements of UDP notes.
- Fix the URL links.
- Remove some extraneous spaces.

### DIFF
--- a/sdk-api-src/content/winsock2/nf-winsock2-wsapoll.md
+++ b/sdk-api-src/content/winsock2/nf-winsock2-wsapoll.md
@@ -260,15 +260,15 @@ When the [WSAPOLLFD](/windows/win32/api/winsock2/ns-winsock2-wsapollfd) structur
 With regard to TCP and UDP sockets:
 
 <ul>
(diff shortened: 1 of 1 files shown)


PRMerger9:
@compnerd : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
I'll approve this in the interest of expediency. We do have many cases such as this, where HTML and markdown are mixed together, and often not co-existing well. However, ideally we'd be moving further toward universal use of markdown, rather than further away from it.

compnerd:
@stevewhims - ah, okay, thats good to know, I'll keep that in mind for future edits.  I tend to run into issues like this often and tend to just put up PRs to at least fix the rendering - it makes reading the docs easier.

stevewhims:
Thanks, @compnerd! Thank you: you definitely fixed the rendering nicely, which is the main priority! If you do change HTML to markdown, ever, the only gotcha really is in Win32 enumeration topics (due to a, hopefully temporary, situation with our content build system), where it's not possible to use markdown within the description of each of the enumerated constants. In that case, we need to stick with HTML in there. Also, if you want to author an emdash in any topic, I use `&mdash;`, which is of course an HTML entity. Apart from that, markdown should work ok everywhere else. :)

Thanks again!
Steve

PR #181, opened 2020-04-07
Correct backslashes in GetTempPathW

ddrinka:
The current documentation is rendered as follows:
> The returned string ends with a backslash, for example, "C:\TEMP".

The second backslash in the example output is applied to the following punctuation character, and should be doubled.


PRMerger7:
@ddrinka : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:
Thanks for your contribution.  I've merged your changes and they should be live soon.

PR #183, opened 2020-04-08
Updated DHCP and DHCPv6 Client API.

fjch1997:
1. Emphasized that _AdapterName_ is a GUID, not the name of the adapter.
2. Added notes that `pclassId` can be null for the DHCPv6 APIs.
3. Added `ERROR_INVALID_NAME` as a possible return value.
4. Added "Or no adapter is found with the GUID specified. " to the return value of `ERROR_INVALID_PARAMETER`.


PRMerger13:
@fjch1997 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #187, opened 2020-04-10
nf-wingdi-getsystempaletteentries.md: nEntries was not defined

halamix2:
nEntries doesn't exist, so it probably should be changed to cEntries


PRMerger20:
@Halamix2 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #191, opened 2020-04-10
Specify LoadLibraryExW dotted filename behavior

jehumb-havok:
Specify the behavior of `LoadLibraryExW()` when passed a module name without a path and `.dll` extension, to highlight the fact that the `.dll` extension is only appended automatically to the module base name if that module base name does not already contain some dot `.` character. This is a source of confusion which leads to a bug in `DllImport` (https://github.com/dotnet/runtime/issues/7223).


PRMerger9:
@djee-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #193, opened 2020-04-13
fix return type

Elite-stay:
"void main" should be "int main"


PRMerger16:
@Elite-stay : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #207, opened 2020-04-16
Clarify IMFMediaBuffer::Lock() and add alternative

jehumb-havok:
Clarify the behavior of IMFMediaBuffer::Lock() regarding memory allocations, and points the reader to better alternatives when the buffer is a 2D buffer supporting either of the IMF2DBuffer and IMF2DBuffer2 interfaces, in order to reduce the chances of the implementation having to internally allocate some memory.

### DIFF
--- a/sdk-api-src/content/mfobjects/nf-mfobjects-imfmediabuffer-lock.md
+++ b/sdk-api-src/content/mfobjects/nf-mfobjects-imfmediabuffer-lock.md
@@ -136,7 +136,7 @@ The pointer returned in <i>ppbBuffer</i> is guaranteed to be valid, and can safe
 
 Locking the buffer does not prevent other threads from calling <b>Lock</b>, so you should not rely on this method to synchronize threads.
 
-This method does not return a newly allocated memory block, nor transfer ownership of the memory to the caller. Do not release or free the memory; the media buffer will free the memory when the media buffer is destroyed.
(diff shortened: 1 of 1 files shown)


PRMerger17:
@djee-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jehumb-havok:
See original discussion on Twitter : https://twitter.com/BruceDawson0xB/status/1250561002019549184

drewbatgit:
Hello. Thanks for the helpful feedback on the documentation. I consulted the developer about this issue and he proposed slightly different language than what you provided. The text has been updated to the following.:

"This method may internally allocate some memory, so if the buffer supports the IMF2DBuffer interface, you should use the IMF2DBuffer::Lock2D method to lock the buffer instead. For 2-D buffers, the Lock2DSize method can be more efficient than the Lock method, depending on the MF2DBuffer_LockFlags value you specify. Calling Lock2DSize with MF2DBuffer_LockFlags_Read won’t incur a copy back when the buffer is unlocked and calling it with MF2DBuffer_LockFlags_Write won’t incur a copy from the internal buffer. Calling Lock2DSize with LockFlags_ReadWrite behaves the same as Lock and Lock2D and will incur a both copy from and copy back when unlocked. The general guidance for best performance is to avoid using IMFMediaBuffer and IMF2DBuffer whenever possible and instead use IMF2DBuffer2 with the minimum required lock flags. Note that if the buffer is locked using Lock2D, the Lock method might return MF_E_INVALIDREQUEST."

PR #208, opened 2020-04-18
Only the first 259 TCHAR's are displayed not 260.

GZPERRA:
Although the list-view control allows any length string to be stored as item text, only the first 259 TCHAR's are displayed. The last character is a null character.


PRMerger10:
@GZPERRA : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #209, opened 2020-04-19
Fix unescaped backslash

loganrosen:
There's an unescaped backslash in `nf-winbase-setcurrentdirectory.md` that leads to nothing being shown instead of the backslash as intended.


PRMerger7:
@loganrosen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #217, opened 2020-04-25
Fixed incorrect inheritance documentation for IAudioClient2

benjarmstrong:
IAudioClient2 is defined in audioclient.h as inheriting from IAudioClient whereas the documentation currently states that it inherits IUknown.


PRMerger6:
@benjarmstrong : Thanks for your contribution! The author(s) have been notified to review your proposed change.

drewbatgit:
Thanks for fixing this. FYI, in the near future we are moving to an automated process that will dynamically insert the correct inheritance information based on reflection of the header.

PR #226, opened 2020-05-02
Propose inner fence documentation for WakeByAddressSingle / WakeByAddressAll

AlexGuteniev:
If there is not such guarantee, then this change is not valid, but then lack of such guarantee should be documented.


PRMerger9:
@AlexGuteniev : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger18:
@AlexGuteniev : Thanks for your contribution! The author(s) have been notified to review your proposed change.

AlexGuteniev:
Is this being reviewed or was abandoned?

Karl-Bridge-Microsoft:
There is pre-existing documentation on this subject, and there is some concern that this statement goes beyond the recommendations.

The references found on this are:

https://docs.microsoft.com/en-us/windows/win32/dlls/run-time-dynamic-linking
https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain
https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices

I’m assuming the reason for the proposed detail is because of the issue where you can’t call LoadLibrary in a DllMain function.

So even if it were true that using GetModuleHandle could be safe in this particular example, we have other suggestions. 

@AlexGuteniev - I'm going to abandon this PR due to the info provided by our dev. If you'd like to instead just add the suggested cross-links, please submit another PR. Thanks!

AlexGuteniev:
Thanks for considering. I don't see cross-links as relevant, so I prefer keeping everything as is then.

AlexGuteniev:
`GetModuleHandle` was not the only goal, inner fence was also proposed, but it is documented here: https://docs.microsoft.com/windows/win32/sync/synchronization-and-multiprocessor-issues#memory-ordering

PR #229, opened 2020-05-04
PageFileUsage and PrivateUsage error

zodiacon:
These counters show the private committed memory (same as Private Bytes in perform or Process Explorer, same as Commit size in Task Manager). It does not include shared committed memory. The text reads: "Total committed memory for a running process" - may be confusing - it is the commit charge for the process, but it's not the total committed memory for the process.


PRMerger16:

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #232, opened 2020-05-04
Edited example to be C++ Standard conformant

b2soft:
Since CreateCommittedResource taking pointers as arguments, old example uses &CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT) call, which returns an instance of D3DX12_HEAP_PROPERTIES).

For address-of operator, "The operand shall be **an lvalue** or a qualified-id.", but "the result of calling a function whose return type is not a reference is a **prvalue**". Current example's code relies on Microsoft's language extension, and is non-conformant to C++ Standard. Thus current code will not compile with **/Za** (disabling C extensions) and **/permissive-** flags.

p.s. Looking forward for other places in documentation where address of non-lvalue is used.

### DIFF
--- a/sdk-api-src/content/d3d12/nf-d3d12-id3d12device-createcommittedresource.md
+++ b/sdk-api-src/content/d3d12/nf-d3d12-id3d12device-createcommittedresource.md
@@ -132,10 +132,12 @@ The <a href="/windows/win32/direct3d12/working-samples">D3D12Bundles</a> sample 
 Create a vertex buffer.
 
 ```cpp
+auto heapProperties = CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT);
+auto resourceDesc = CD3DX12_RESOURCE_DESC::Buffer(SampleAssets::VertexDataSize);
 ThrowIfFailed(m_device->CreateCommittedResource(
-    &CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT),
+    &heapProperties,
     D3D12_HEAP_FLAG_NONE,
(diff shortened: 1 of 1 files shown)


PRMerger12:
@b2soft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thanks for the change, it's definitely better. For the original version of the code, I dug out a project and introduced this error into it. The project already had /permissive-, and it built. I then (as you say) set /Za, but then I got all kinds of errors (about __super and __uuidof, for example; and I need those in that project). So those errors masked any error about taking the address of this prvalue. So I suspect that anyone using features like I was will be forced to use extensions anyway and so this issue will be masked. All that said, your change is the way I'd hope folks would do this. Thanks again!

b2soft:
> 
> 
> Thanks for the change, it's definitely better. For the original version of the code, I dug out a project and introduced this error into it. The project already had /permissive-, and it built. I then (as you say) set /Za, but then I got all kinds of errors (about __super and __uuidof, for example; and I need those in that project). So those errors masked any error about taking the address of this prvalue. So I suspect that anyone using features like I was will be forced to use extensions anyway and so this issue will be masked. All that said, your change is the way I'd hope folks would do this. Thanks again!

I posted a bug to MSVC about /permissive- is not enough to catch such address-of violations:
https://developercommunity.visualstudio.com/content/problem/1015619/taking-address-of-a-temporary-is-not-catched-by-th.html
Thus, _probably_ if MSVC is fixed, such code will break with /permissive- (since /Za is not related to C++)

I will dig into docs more and post other PRs if any (about conformance with prvalues).

PR #238, opened 2020-05-05
Update ns-rectypes-lattice_metrics.md

035:
Changed the word `Height` to `Offset` as `Height` implies a nonnegative value, but this value can be negative as it is an offset.


PRMerger15:
@035 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #243, opened 2020-05-08
Missing and errornous info

zodiacon:
The process handle should have PROCESS_VM_OPERATION access mask
The import library seems to be micore.lib (not kernel32.lib)


PRMerger8:

Karl-Bridge-Microsoft (changes requested):
Please revert the req.lib revision. Every other topic in that docset has the same detail, so we need to be consistent across all related topics. You can file a Git Issue for us to track and create a more extensive work item for that specific task if required.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #247, opened 2020-05-11
Add reference for hSource field

ALaborate:
Add reference to examples for hSource field description. 

It was misleading to me and that cost a while. A mention that further info is provided in an example section below would have been helpful.

Not sure about `^M` deletion. I haven't done it intentionally and it seem to me that it was done by github web interface which I used to create commit. Should I remove it manually, or is it acceptable?


PRMerger17:
@OMOH98 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #258, opened 2020-05-14
Fix check for relative movement in RAWMOUSE sample code

andrei-drexler:
Since MOUSE_MOVE_RELATIVE is 0, and-ing it with any integer also returns 0, making the original condition always true. Proposed change first checks for absolute mouse movement, handling relative movement in the else branch.


PRMerger19:
@andrei-drexler : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ChrisGuzak (approved):

jwmsft (approved):

PR #275, opened 2020-05-22
Task: 221254 Broken link fixes

atikmapari:


stevewhims:
For the first change, I noticed that the url is absolute. I'd recommend replacing all `https://docs.microsoft.com/` in links with `/`. That's to support the Microsoft Help Viewer.

After that, I found conducting the rest of the review impractical as it stands. Unfortunately, because the very first change in many files is the automatic conversion of a newline char, the remainder of the diff in those cases makes it appear that the entire file has changed. That makes it impossible to see the actual change(s). So that issue compounded with the fact that 71 files have been changed in this PR (which is a lot) makes this PR really difficult to verify.

I can approve it if necessary, but unfortunately the way it is I can't verify the links.

Alternatively, if you prefer, you could take a first pass through all the files making just any old change (for example adding a newline to the end of the file), and then I'd be happy to approve that if you describe what you've done. After that take another pass through, say, 10 files at a time making the actual link fixes. That would solve these two issues: a) it'll make the actual, real changes to the files show up in the diff, and b) it'll be more manageable for approvers signing off on the changes.

Thanks!
Steve

drewbatgit (approved):

atikmapari:
Hi @stevewhims
1. All "https://docs.microsoft.com/" links have been replaced with " / " in updated 71 files . 
2. In previous changes in meta tags "description" and "helpviewer_keywords" was in one line and in latest updated files "description" and "helpviewer_keywords" is coming in separate lines. Hence in PR files changes section it's showing all contents in files updated.
 
Please let me know, If you want to make any changes in all 71 pages regarding 2nd point..

PR #280, opened 2020-05-26
Fixed bad links

jasonepperly:
File, power, and volume control codes were missing initial forward slash. This resulted on those 3 links hitting a 404.


PRMerger18:
@jasonepperly : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #305, opened 2020-06-09
Fix error regarding SetClipboardData in response to WM_RENDERALLFORMATS

metathinker:
When a window responds to the  [`WM_RENDERALLFORMATS` window message](https://docs.microsoft.com/en-us/windows/win32/dataxchg/wm-renderallformats), it needs to call OpenClipboard before calling SetClipboardData; otherwise its SetClipboardData calls will fail and the data won't be set.

This change removes an incorrect contrary claim in [the documentation for SetClipboardData](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclipboarddata), and updates a bit of markup on the same page from HTML to Markdown.

Pairs with https://github.com/MicrosoftDocs/win32/pull/397

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-setclipboarddata.md
+++ b/sdk-api-src/content/winuser/nf-winuser-setclipboarddata.md
@@ -60,7 +60,7 @@ ms.custom: 19H1
 ## -description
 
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@metathinker : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PRMerger14:
@jwmsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #307, opened 2020-06-10
Clarify lifetime of handle returned by IDXGISwapChain2::GetFrameLaten…

beviu:
…cyWaitableObject

From the documentation, it is not very clear whether the handle that is returned is owned by the swap chain and should not be closed or if it should be closed by the application.

When I looked at this page, I clicked on the "DirectXLatency sample" link, read the sample and because the sample does not close the handle, I assumed that I shouldn't close the handle:
https://github.com/microsoftarchive/msdn-code-gallery-microsoft/tree/master/Official%20Windows%20Platform%20Sample/DirectX%20latency%20sample/%5BC%2B%2B%5D-DirectX%20latency%20sample/C%2B%2B

However, you should close the handle:
https://github.com/microsoft/terminal/pull/6435#discussion_r438303247


PRMerger14:
@greg904 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #328, opened 2020-06-20
Add example drive path.

AustinWise:


PRMerger17:
@AustinWise : Thanks for your contribution! The author(s) have been notified to review your proposed change.

AustinWise:
Using quotes like the other examples looked like this: "\\?\PhysicalDrive2". At least on GitHub, you can see that the double backslash is treated like an escape character. Since I don't know if it will look the same when published, I used the other quote style.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #332, opened 2020-06-23
Update return of ExtractIconEx

dkd1111:
the return case is added.

### DIFF
--- a/sdk-api-src/content/shellapi/nf-shellapi-extracticonexa.md
+++ b/sdk-api-src/content/shellapi/nf-shellapi-extracticonexa.md
@@ -109,7 +109,7 @@ The number of icons to extract from the file.
 
 Type: <b>UINT</b>
 
-If the <i>nIconIndex</i> parameter is -1, the <i>phiconLarge</i> parameter is <b>NULL</b>, and the <i>phiconSmall</i> parameter is <b>NULL</b>, then the return value is the number of icons contained in the specified file. Otherwise, the return value is the number of icons successfully extracted from the file.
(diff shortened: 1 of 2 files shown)


PRMerger13:
@dkd1111 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger10:
@dkd1111 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@dkd1111 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

dkd1111:
https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/df7946b2-999c-4684-a740-4712f5ad6e17/return-of-extracticonex?forum=windowssdk#df7946b2-999c-4684-a740-4712f5ad6e17

mcleanbyron:
Thanks for your contribution. As per the discussion on the forum post you linked to, these functions return the UINT value -1 if they encounter an error (for example, if the specified file can't be found). This is pretty common behavior for Win32 APIs. Given this, I am closing this pull request, and I am adding the following details to the "Return value" sections. This new info should be published to the live docs in the next day.

"If the file specified by *lpszFile* cannot be found or some other error occurs, this function returns -1 and you can call [GetLastError](https://docs.microsoft.com/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror) to retrieve the error code."

dkd1111:
It's not just "NOT found error".

ExtractIconEx(L"notfoundfile", -1, nullptr, nullptr, 0);
This code return 0. I think it's right because it doesn't extract any icon.
The case written in my commit returns -1.
Additionally, you should not write "return -1" because of its return type.

 Given that the number of icons contained in the specified file is 429496729, it is considered to have a bug.

mcleanbyron:
Thanks for your reply, and good point that the actual text in the article should not indicate the return value is -1 given the UINT type. UINT_MAX/4,294,967,295 is the UINT equivalent to 0xFFFFFFFF (hex) and -1 (int). I'm planning to add the following note to the return values to clearly indicate that this is an error condition for the specific case you originally called out in your proposed update and for potentially other error conditions too.
_______

**Note** If the function encounters an error, it returns **UINT_MAX**. In this case, you can call [GetLastError](https://docs.microsoft.com/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror) to retrieve the error code. For example, this function returns **UINT_MAX** if the file specified by *lpszFile* cannot be found while the *nIconIndex* parameter is any value other than -1 and either *phiconLarge* or *phiconSmall* is not **NULL**. In this case, **GetLastError** returns **ERROR_FILE_NOT_FOUND** (2).

PR #361, opened 2020-07-09
Fix links in D3D12_RENDER_PASS_BEGINNING_ACCESS

john-h-k:


PRMerger16:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (changes requested):
It's important not to include the language tag in urls (for the benefit of any customer not using a browser set to en-us). So, for example it's important not to include the `/en-us/` in `https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_render_pass_beginning_access_type`.

After that, it's also a good idea to transform any docs.microsoft.com url into a site-relative one. That's so that the url works not just online but also in the Microsoft Help Viewer. That's easily done by taking a pass replacing all `https://docs.microsoft.com/` with `/`. The result of this step and the previous one is to be left with urls of the form `/windows/win32/api/d3d12/ne-d3d12-d3d12_render_pass_beginning_access_type`. Those are site-relative urls.

Thanks!

PRMerger13:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

john-h-k:
Ah, remebered to skip the language tag for 2 of them but forgot one 😄 , fixed now

stevewhims (approved):
Excellent, thank you!

PRMerger10:

PR #368, opened 2020-07-16
Fix reference to directory type enum value

jasonknichel:


hkhmsft:
Appreciate the contribution, @jasonknichel .

PR #370, opened 2020-07-16
Separate WinHTTP and HTTP.SYS categories

Techwolfy:
The documentation for WinHTTP and HTTP.SYS is combined into a single "Windows HTTP Services (WinHTTP)" category on MSDN, which causes some confusion. This PR splits the client and server documentation into separate categories.

### DIFF
--- a/sdk-api-src/content/_http/config.json
+++ b/sdk-api-src/content/_http/config.json
@@ -4,7 +4,6 @@
     "httpserv.h",
     "http.h",
     "httpp.h",
-    "winhttp.h"
   ],
-  "displayName": "Windows HTTP Services (WinHTTP)"
+  "displayName": "HTTP Server API"
 }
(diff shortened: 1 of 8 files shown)


PRMerger14:
@Techwolfy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Techwolfy:
@stevewhims, please take a look and make sure I did this the right way; I had to create a new GUID for the new category file and I'm not sure how the content publishing system handles that.

stevewhims:
Note that we're not using MSDN any longer (we've moved to docs.microsoft.com, or DMC for short). AFAIK, `ms.assetid:` is used only for topics *that were ever published on MSDN*. Since these new files are new for DMC, they were never published on MSDN, so I think we'll be fine. No need to remove the `ms.assetid:`; it's redundant, but it won't do any harm.

Also, just FYI, "categories" are known as "technologies". So here you've created a new technology named "winhttp". It's also worth knowing that each API reference topic contains yaml metadata indicating which "technology" that API is part of. So to indicate that a given API is part of the "winhttp" technology, we'll need to add the metadata `tech.root: winhttp` to topics for all APIs that are part of that tech.

stevewhims (approved):
Looks good. See my comment about moving APIs into this new technology.

PR #374, opened 2020-07-21
Make sure __cdecl is not missed.

pmsjt:
Add an explicit "syntax" section. so make sure the calling-convention keyword (__cdecl) is not missed.

### DIFF
--- a/sdk-api-src/content/winnt/nf-winnt-rtlrestorecontext.md
+++ b/sdk-api-src/content/winnt/nf-winnt-rtlrestorecontext.md
@@ -56,6 +56,14 @@ ms.custom: 19H1
 
 Restores the context of the caller to the specified context record.
 
+## -syntax
+
+```cpp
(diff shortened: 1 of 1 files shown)


PRMerger20:
@pmsjt : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, just a heads up that we (content) ought to track the work to revert this after https://ceapex.visualstudio.com/Engineering/_workitems/edit/223683 is fixed.

-Steve

pmsjt:
> Hi, just a heads up that we (content) ought to track the work to revert this after https://ceapex.visualstudio.com/Engineering/_workitems/edit/223683 is fixed.
> 
> -Steve

Hi Steve.

Just a small note: The aforementioned bugs states "callback calling conventions". This is too restrictive and should be replaced by "calling conventions", As we can see by this example, the calling convention omissions are not particular to callbacks.

PR #384, opened 2020-07-31
Clarify documentation for PrjMarkDirectoryAsPlaceholder

xavierd:
The targetPathName description is a bit misleading, and actually expects a full path, and not a relative path. Attempting to give it a relative path will always result in HRESULT(ERROR_INVALID_PARAMETER).


PRMerger14:
@xavierd : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #393, opened 2020-08-07
Add documentation for 'ID3DDestructionNotifier'

john-h-k:
Type has been present in `D3DCommon.h` for a while but isn't publicly documented

### DIFF
--- a/sdk-api-src/content/d3dcommon/nn-d3dcommon-id3ddestructionotifier.md
+++ b/sdk-api-src/content/d3dcommon/nn-d3dcommon-id3ddestructionotifier.md
@@ -1,0 +1,108 @@
+---
+UID: NN:d3dcommon.ID3DDestructionNotifier
+title: ID3DDestructionNotifier (d3dcommon.h)
+description: ID3DDestructionNotifier is an include interface that the user implements to allow an application to call user-overridable methods for opening and closing shader
+helpviewer_keywords: ["ID3DDestructionNotifier","ID3DDestructionNotifier interface [Direct3D]","ID3DDestructionNotifier interface [Direct3D]","described","d3dcommon/ID3DDestructionNotifier","direct3d.id3ddestructionnotifier"]
+old-location: direct3d11\id3dinclude.htm
+tech.root: direct3d11
+ms.assetid: 2020ce65-3a6e-4a9f-9e97-b94e3c75f4f5
+ms.date: 12/05/2018
+ms.keywords: ID3DDestructionNotifier, ID3DDestructionNotifier interface [Direct3D], ID3DDestructionNotifier interface [Direct3D],described, d3dcommon/ID3DDestructionNotifier, direct3d.id3ddestructionnotifier
+f1_keywords:
+- d3dcommon/ID3DDestructionNotifier
+dev_langs:
+- c++
+req.header: d3dcommon.h
+req.include-header: 
+req.target-type: Windows
+req.target-min-winverclnt: Windows 7 [desktop apps \| UWP apps]
+req.target-min-winversvr: Windows Server 2008 R2 [desktop apps \| UWP apps]
+req.kmdf-ver: 
+req.umdf-ver: 
+req.ddi-compliance: 
+req.unicode-ansi: 
+req.idl: 
+req.max-support: 
+req.namespace: 
+req.assembly: 
+req.type-library: 
+req.irql: 
+topic_type:
+- APIRef
+- kbSyntax
+api_type:
+- COM
+api_location:
+- D3DCommon.h
+api_name:
+- ID3DDestructionNotifier
+targetos: Windows
+req.typenames: 
+req.redist: 
+ms.custom: 19H1
+---
+
+# ID3DInclude interface
+
+
+## -description
+
+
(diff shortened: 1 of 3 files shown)


PRMerger15:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

sdk-api-src/content/d3dcommon/nn-d3dcommon-id3ddestructionotifier.md
```diff
@@ -0,0 +1,108 @@
...
+## -description
+
+
+<b>ID3DDestructionNotifier</b> is an interface that can be used to register for callbacks when a D3D Nano-COM object is destroyed.
+To acquire an instance of this interface, call <a href="windows/win32/api/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void)"></a> on a D3D object with the **IID** of **ID3DDestructionNotifier**.
+
+
+## -inheritance
+
+The <b xmlns:loc="http://microsoft.com/wdcml/l10n">ID3DDestructionNotifier</b> interface inherits from the <a href="https://docs.microsoft.com/windows/desktop/api/unknwn/nn-unknwn-iunknown">IUnknown</a> interface. <b>ID3DInclude</b> also has these types of members:
```
jenatali: Typo - ID3DInclude?

sdk-api-src/content/d3dcommon/nn-d3dcommon-id3ddestructionotifier.md
```diff
@@ -0,0 +1,108 @@
...
+Unregisters a user-defined callback that was previously registered.
+
+</td>
+</tr>
+</table> 
+
+
+## -remarks
+
+The <b>ID3DDestructionNotifier</b> can be used to track resources which are being unexpectedly released early, or providing a log of object disposal.
```
jenatali, line 97: Note that destruction can also be tracked by using `ID3D11DeviceChild::SetPrivateDataInterface` or `ID3D12Object::SetPrivateDataInterface`, and using the provided object's `Release()` implementation to track destruction. The key difference between that object and this is ordering of destruction for intertwined objects.

For example, a D3D11 view holds an internal reference on a D3D11 resource. Using the `SetPrivateDataInterface` approach, if the resource is only being kept alive by the view, then releasing the view will result in the resource being destroyed first (as the view's internal reference to the resource is released before destroying the private data container).

The destruction notifier is specifically designed to explicitly order destruction of objects so that immutable object relationships, such as a view -> resource, are still valid at the time the notifiers are invoked.

jenatali (changes requested):

PRMerger17:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

sdk-api-src/content/d3dcommon/nn-d3dcommon-id3ddestructionotifier.md
john-h-k: Oops, copied the structure of this from that (as I have no idea what I am doing wrt writing docs). Will fix

PRMerger20:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jenatali (approved):

stevewhims:
Hi, @john-h-k , and thanks for filling gaps in the docs by adding these topic files! :)

I'll take the PR as-is, and I've created a work item for myself to come back and fix up the topics, and also to do the remaining work to bring the relevant TOCs and landing pages (in the conceptual repo) up to date with these new topics. None of the changes need prevent us from going ahead with this PR, though.

If you're curious, I can touch on the kinds of changes I'll make to the topics.

It looks like you based the new topics on existing ones. There is a gap though between how topics *should* be authored, and how a lot of existing topics actually *are* authored (probably most of them, I'd say). So these are just issues we learn about by experience, and try to remove them from old docs if we can, and try not to let them get into new topics if possible.

Since topics for these APIs were never published (as far as I know), several yaml fields are not needed. old-location, ms.assetid, for example. In Win32 API ref, there should be no H1 in the topic (the first heading with a single hash). We'd want to link to types (for example, **ID3DDestructionNotifier**), at least on first mention, rather than just bold them. The value proposition of the "Type:" headings is to link to the type in question (and sometimes to show the SAL). So if we omit a hyperlink then there's arguably no value to the "Type:" section, since the syntax block already tells you the type but has no link. The `## -members` section is redundant, and we'd want to remove that from the topic. The `## -inheritance` section is slightly malformed. Again, as I say, many of these problems exist in old, existing topics, so if you've copied from one then it's to be expected that they'll be here, too. There are a few sentences with no period at the end, and I'll do other low-pri stylistic/formatting tweaks. :)

For the stuff I mentioned about the conceptual repo, that'd be adding links into certain TOCs and landing pages in the Direct3D 11 reference (https://docs.microsoft.com/windows/win32/direct3d11/d3d11-graphics-reference). Each topic in the API repo is represented in those places in TOCs and landing pages in the reference section in the conceptual repo.

Again, thanks for contributing, and don't worry about the tidy-up I describe above: I'll get to it when I can. Meanwhile, these topics resolve the problem that we didn't have them before.

Cheers!
Steve

PR #399, opened 2020-08-11
winreg.h/advapi32.dll: State that RegGetValue accepts null subkey name

metathinker:
Like the older [SHRegGetValue](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shreggetvaluea
) function on which it is based, [RegGetValue](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluea) allows you to specify a null reference or empty string for the `lpSubkey` parameter, which means to query the base key referred to by the `hkey` parameter.

This change adds that piece of information to the RegGetValue documentation pages.

### DIFF
--- a/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
+++ b/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
@@ -93,24 +93,21 @@ This handle is returned by the
 
 ### -param lpSubKey [in, optional]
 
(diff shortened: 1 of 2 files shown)


PRMerger16:
@metathinker : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #401, opened 2020-08-12
Update ns-winuser-rawinputdevice.md

rapsealk:
I found that **RIDEV_CAPTUREMOUSE** and **RIDEV_NOHOTKEYS** have the same value(`0x00000200`).
I'm not sure whether it is correct, so I'm asking for a re-check.

Thank you.


PRMerger12:
@rapsealk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #413, opened 2020-08-18
Change example locale from en-FJ to en-XA for ResolveLocaleName

erik0686:
This PR removes "en-FJ" being used as an example of a fake or non existing locale. 

It seems that when [this](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-resolvelocalename) doc file was created, en-FJ did not exist, but now it does so it doesn't make sense having it being used as an example of a fake locale being resolved into an existent one. 

@jefgen suggested that using that using a "Private Use" ISO region code might be better, as it can't be used for a real locale, and makes the example more obvious that it is a constructed locale

I did a search on the repo using vs code and the GitHub search and it seems this is the only file that uses en-FJ as an example, but it wouldn't hurt to keep an eye on other files that might be having the same problem.

### DIFF
--- a/sdk-api-src/content/winnls/nf-winnls-resolvelocalename.md
+++ b/sdk-api-src/content/winnls/nf-winnls-resolvelocalename.md
@@ -67,12 +67,12 @@ Finds a possible <a href="https://docs.microsoft.com/windows/desktop/Intl/locale
 
 ### -param lpNameToResolve [in, optional]
 
-Pointer to a name to resolve, for example, "en-FJ" for English (Fiji).
+Pointer to a name to resolve, for example, "en-XA" for English (Private Use).
 
 
 ### -param lpLocaleName [out, optional]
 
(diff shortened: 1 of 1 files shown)


PRMerger16:
@erik0686 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jefgen:
Thanks @erik0686!

> I did a search on the repo using vs code and the GitHub search and it seems this is the only file that uses en-FJ as an example,

Thanks for checking! 👍 

Also tagging others as a FYI, since I can't add them to the PR, or edit the main PR description:
 @ShawnSteele, @axelandrejs, @microsoft/target-dev

jefgen (approved):
LGTM, thanks for fixing this!

jefgen:
nit: I'd suggest re-wording the PR title to be something like this in order to be more clear as to the change:
"Change example locale from en-FJ to en-XA for ResolveLocaleName"

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #434, opened 2020-08-29
updating req.lib

ff-k:
Applications using this function should be linked against mfplat.lib. Tested on Windows 8 with MSVC 19 for x64.


PRMerger8:
@ff-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #437, opened 2020-09-01
Correct information about setting the system time zone

mattjohnsonpint:
The doc is incorrect, stating that the dynamic call is for Windows 7 and 8 only.  It's actually for Windows Vista / Server 2008 and newer, including through Windows 10.

Note: It would also be good if the `SetDynamicTimeZoneInformation` page showed an example of setting the time zone by enumerating zones with `EnumDynamicTimeZoneInformation` and choosing one that matched the key name - along with setting the required permissions, etc.  I don't have time to include that today, but I wanted to get this other change in as it's simply incorrect.

Thanks.


PRMerger7:
@mj1856 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

PR #446, opened 2020-09-11
EnterCriticalSection: Note about processes currently exiting

Dwedit:
According to https://devblogs.microsoft.com/oldnewthing/20100122-00/?p=15193 (and observed behavior), a call to EnterCriticalSection that blocks will terminate if the process is in the middle of exiting.


PRMerger17:
@Dwedit : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #456, opened 2020-09-16
Update nf-processthreadsapi-getcurrentthreadeffectivetoken.md

GovM-MSFT:
Adding min supported OS as windows 8

### DIFF
--- a/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-getcurrentthreadeffectivetoken.md
+++ b/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-getcurrentthreadeffectivetoken.md
@@ -11,8 +11,8 @@ ms.keywords: GetCurrentThreadEffectiveToken, GetCurrentThreadEffectiveToken func
 req.header: processthreadsapi.h
 req.include-header: 
 req.target-type: Windows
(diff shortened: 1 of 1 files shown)


PRMerger19:
@GovM-MSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

GovM-MSFT:
@Karl-Bridge-Microsoft  - Thanks for the approval... but it appears I can't complete the merge, so what is the next step?

Karl-Bridge-Microsoft:
Hi @GovM-MSFT, I approved and then thought I should look into this further as the info should be in the topics already. Hence the delay. 

I confirmed that the info should be in the topic, but that it might've beev dropped during a docs migration a few years ago.

Only writers can finalize a PR, so I'll do that now.

GovM-MSFT:
Oh great, thank you very much. BTW, this is the process I prefer as well. We make the suggestions, and then you guys make the final determination, by completing the PR.


Gov (Rhymes with Orange)

Work It Harder Make It Better
Do It Faster, Makes Us stronger
More Than Ever, Hour After Hour
Work Is Never Over

________________________________

PR #458, opened 2020-09-16
Add clarification about IO completion ports

mucki-at:
Add short paragraph to clarify that IO completion ports work as expected when the read completes immediately. 

NOTE: if accepted, this change should be duplicated to all similar methods (WSASendTo, ...) but I don't have a list. If you provide me with one I can make the changes.


PRMerger8:
@mucki-at : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, @mucki-at, I think this is the list that you mention in your description: https://docs.microsoft.com/windows/win32/fileio/i-o-completion-ports#supported-io-functions.

stevewhims (approved):
Looks good! (FYI, I think this is the list that you mention in your description: https://docs.microsoft.com/windows/win32/fileio/i-o-completion-ports#supported-io-functions.)

PR #461, opened 2020-09-18
ClientAddress not available in some scenarios

gbrag-ms:
The details are documented here:
Bug 29439207: The function WTSQuerySessionInformation does not return the ClientAddress if the connection has been established through the RDS Gateway
Bug 29464971: The function WTSQuerySessionInformation does not return the ClientAddress if the connection originated from the "Microsoft Remote Desktop" Store App


PRMerger16:
@gbrag-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #467, opened 2020-09-18
Clarify this function requires filesystem support

kyanha:
This function relies on filesystem support. If a filesystem does not support it, ERROR_INVALID_PARAMETER is returned.

Also, the type of the streams this function affects can best be described here as ':$DATA' (technically NTFS defines ':' as the delimiter and '$DATA' as the stream type, for the final construction of `filename:streamname:streamtype`; for those who don't work much with filesystems, though, it seems a rather academic distinction).  '::$DATA' refers to the unnamed (default) data stream on a file.  Directories cannot have an unnamed data stream, so referring to "the first ::$DATA stream type" is incorrect.

Thanks for working as hard as you do on making sure we can understand what your company has created for us to use!


PRMerger15:
@kyanha : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:
Hi @kyanha! Thanks for the contribution. It looks like there's a merge conflict - once this is addressed, I'd be happy to merge these improvements.

PRMerger19:
@kyanha : Thanks for your contribution! The author(s) have been notified to review your proposed change.

kyanha:
I also changed a couple of absolute URLs to be relative.

MatchaMatch:

PR #483, opened 2020-10-01
Fix name requirement for NamedPipe

Krzmbrzl:
The double backslash in the markdown was interpreted as a quoted single backslash which would
cause the documentation to state that the name must be `\.\pipe` instead of `\\.\pipe`


PRMerger16:
@Krzmbrzl : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #514, opened 2020-10-23
Update nf-winuser-settimer.md

wotent:
DispatchMessage function: If the lpmsg parameter points to a WM_TIMER message and the lParam parameter of the WM_TIMER message is not NULL, lParam points to a function that is called instead of the window procedure.


PRMerger20:
@wotent : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #526, opened 2020-10-30
Remove statement that SleepEx will return early

vtjnash:
I'm proposing deleting the following text:

> If dwMilliseconds is less than the resolution of the system clock, the thread may sleep for less than the specified length of time. If dwMilliseconds is greater than one tick but less than two, the wait can be anywhere between one and two ticks, and so on.

Since empirically when I measure the time taken by `Sleep`, this doesn't seem true, but I would like confirmation from the engineers at Microsoft.

If it's confirmed that this is not the implementation, the same correction is also needed at https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep and https://docs.microsoft.com/en-us/windows/win32/sync/wait-functions#wait-functions-and-time-out-intervals


PRMerger7:
@vtjnash : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

vtjnash:
Thanks for confirming!

Karl-Bridge-Microsoft:
Hi @vtjnash - I'm still waiting for confirmation from the dev team. I'll update the PR as soon as i get that.

vtjnash:
bump?

Karl-Bridge-Microsoft:
Confirmation still pending.

vtjnash:
Bump?

Karl-Bridge-Microsoft:
Hi @vtjnash - A dev from the feature team replied and confirmed your findings and the revisions to the doc in your PR. Thanks for your patience!

Karl-Bridge-Microsoft:
Would you like to update the associated docs you specify in another PR?

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
#sign-off

vtjnash:
No thanks. Very much appreciate the confirmation though!

PR #533, opened 2020-11-05
Proposed additions to WTS_CONNECTSTATE_CLASS

muiriswoulfe:
While working with `WTS_CONNECTSTATE_CLASS` recently, both I and a colleague had difficulty interpreting all of the possible enumeration values based on this documentation. Through experimentation, we ascertained the information I've now added via this PR.

I think it would be great to add these clarifications to the documentation, but I also have some reservations, so I'm really sending this for a discussion. In particular, I'm not 100% satisfied with the following:

- I can't tell if the information is complete. I might have uncovered certain pieces of information, while inadvertently excluding others.
- The details are only being added for the two values I examined. It would be great to include similar details for the other values.

Ideally, it would be great if we could loop a Subject Matter Expert into this review as well.

I'm happy to make whatever changes we decide on, or to shelve this review entirely if favoured.

### DIFF
--- a/sdk-api-src/content/wtsapi32/ne-wtsapi32-wts_connectstate_class.md
+++ b/sdk-api-src/content/wtsapi32/ne-wtsapi32-wts_connectstate_class.md
@@ -60,6 +60,8 @@ Specifies the connection state of a Remote Desktop Services session.
 
 A user is logged on to the WinStation.
 
+This state occurs when a user is signed in and actively connected to the device.
+
 ### -field WTSConnected
 
 The WinStation is connected to the client.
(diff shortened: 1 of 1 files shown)


PRMerger10:
@muiriswoulfe : Thanks for your contribution! The author(s) have been notified to review your proposed change.

muiriswoulfe:
FYI, I didn't make any changes to the last line of the file and I used the GitHub editor for making these changes, so I wouldn't have anticipated any change like this. My guess is it is a line-ending change that should have no impact.

Let me know if something needs to be fixed there too.

mcleanbyron (approved):

PRMerger19:

muiriswoulfe:
@mcleanbyron Any update on this PR? I'm happy to close it if desired.

mcleanbyron:
It will be very difficult to find an internal subject matter expert to review these changes. However, given the relatively small scope of these changes, I'm comfortable with merging them. Thanks for your contribution!

PR #544, opened 2020-11-12
Update nf-wuapi-iupdate-get_image.md

arcarley:
Under Return Value it should state:
"Returns S_OK if successful. Otherwise, returns a COM or Windows error code."
Under Remarks, it should state:
"This API can return a null pointer as the output, even when the return value is S_OK."

(You may need to reformat this as these changes look strange in the preview.. perhaps this is automated?)


PRMerger18:
@arcarley : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:

mcleanbyron (approved):

PR #549, opened 2020-11-19
D3D12 Documentation minor updates

ScatteredRay:
I've been freshly working with D3D12 lately, and while doing so, there were some aspects that tripped me up a bit.

I've made some suggestions to the docs that would clarify some confusing/missing elements that I encountered. And would have vastly speed up my time to initial implementation.

Mostly minor things, but I think they would be super valuable.


PRMerger13:
@ScatteredRay : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #564, opened 2020-11-24
Remove note about MAX_PATH limitation

justanotheranonymoususer:
This limitation doesn't really exist, and just confuses as of whether these functions can be used with long paths - they can.

### DIFF
--- a/sdk-api-src/content/shlwapi/nf-shlwapi-pathmatchspeca.md
+++ b/sdk-api-src/content/shlwapi/nf-shlwapi-pathmatchspeca.md
@@ -63,13 +63,13 @@ Searches a string using a Microsoft MS-DOS wildcard match type.
 
 Type: <b>LPCSTR</b>
 
-A pointer to a null-terminated string of maximum length MAX_PATH that contains the path to be searched.
+A pointer to a null-terminated string that contains the path to be searched.
 
 ### -param pszSpec [in]
 
 Type: <b>LPCSTR</b>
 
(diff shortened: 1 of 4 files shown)


PRMerger14:

PRMerger12:

PRMerger12:

mcleanbyron (approved):

PRMerger12:

mcleanbyron:
Thanks for your changes. I agree that the MAX_PATH requirement applies to fewer scenarios in recent versions, but I added a note to the relevant article in the remarks.

justanotheranonymoususer:
> I added a note to the relevant article in the remarks

I don't have strong feelings about this, but I don't see how the note clarifies the situation. The functions, or shlwapi in general, are not mentioned in the linked article. Also, the functions operate on a stringish path, no on the filesystem, and the string operation doesn't care about the limitations of FAT, NTFS, and so on. So I think that this note just makes things more confusing. I'd rather say there's no limit at all, but I'm not sure whether such limits existed in old Windows versions.

mcleanbyron:
Thanks for your additional feedback! Based on previous experience with similar APIs, I made the assumption that the path length limitation statements that were in these articles referred to internal operations utilized by these functions, even though the MAX_PATH article I linked to didn't refer to these functions specifically. However, based on your reply I took a closer look at the current implementations of these functions and don't see any obvious checks for MAX_PATH. I removed the remarks about MAX_PATH applying in some environments (perhaps this check used to be there in an older release...). This latest doc change should go live in the next day or two. 

Thanks again!

PR #565, opened 2020-11-24
Add notes about amplification and mesh shaders

pyjarrett:
I'm hoping this is auto-generated from code, so these probably should be underlying code changes.


PRMerger18:
@pyjarrett : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

stevewhims:
This is the correct change. FYI, the `### -field Xxx` part is auto-gen'd into a stub. We then have to diff stubs/topics and reconcile. No underlying code changes necessary. Thanks!

PR #570, opened 2020-11-29
LookupAccountSidA: ERROR_INSUFFICIENT_BUFFER

skissane:
Documentation did not mention that ERROR_INSUFFICIENT_BUFFER is the error code when buffer is too small. Although that would be obvious to anyone familiar with Win32 APIs, mentioning it explicitly will help avoid any doubt and help newcomers.

Also mention that you are allow to pass NULL pointers for the buffers if the buffer size is being passed as zero.

### DIFF
--- a/sdk-api-src/content/winbase/nf-winbase-lookupaccountsida.md
+++ b/sdk-api-src/content/winbase/nf-winbase-lookupaccountsida.md
@@ -73,15 +73,17 @@ A pointer to the
 
 ### -param Name [out, optional]
 
-A pointer to a buffer that receives a <b>null</b>-terminated string that contains the account name that corresponds to the <i>lpSid</i> parameter.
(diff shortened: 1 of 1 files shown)


PRMerger7:
@skissane : Thanks for your contribution! The author(s) have been notified to review your proposed change.

code-creator:
If a win api function returns success, it makes IMHO no sense to document the GetLastError() value.
If one wants to get the size of the buffer, and the function returns true it is no legitimate reason to call GetLastError().
As the doc for GetLastError() states, it may contain an unknown value if last win api call returns success.
But here either the function is either an exception to other win api functions, or the documentation is incorrect. In other win api functions, if one passes zero as the buffer size to get the required buffer size they return success if they can fill in the required buffer size. For this function, the documentation states it returns failure in this case. If this is correct, the proposed change to the documentation would be legitimate.

PR #577, opened 2020-12-01
API argument order is wrong

SpriteOvO:
First, please do not merge this commit directly.
Since the repository does not enable `GitHub issue tracker`, I can only open a `PR` to report the error.

---

This is the function declaration for `CreateTimerQueueTimer` in SDK 10.0.18362.0.

```cpp
WINBASEAPI
BOOL
WINAPI
CreateTimerQueueTimer(
    _Outptr_ PHANDLE phNewTimer,
    _In_opt_ HANDLE TimerQueue,
    _In_ WAITORTIMERCALLBACK Callback,
    _In_opt_ PVOID Parameter,
    _In_ DWORD DueTime,
    _In_ DWORD Period,
    _In_ ULONG Flags
    );
```
In its [MSDN page](https://docs.microsoft.com/en-us/windows/win32/api/threadpoollegacyapiset/nf-threadpoollegacyapiset-createtimerqueuetimer), the `Parameter` argument is incorrectly displayed at the end.

image
image

But it is in the correct order in the repository.
So I guess it was a mistake in the `Syntax` heading, but the `Syntax` heading is probably in an undisclosed place, because I didn't find them. Am I missing something?

Please correct me if I'm wrong.
Thanks!

### DIFF
--- a/sdk-api-src/content/threadpoollegacyapiset/nf-threadpoollegacyapiset-createtimerqueuetimer.md
+++ b/sdk-api-src/content/threadpoollegacyapiset/nf-threadpoollegacyapiset-createtimerqueuetimer.md
@@ -50,7 +50,7 @@ api_name:
  - CreateTimerQueueTimer
 ---
 
-# CreateTimerQueueTimer function
+# CreateTimerQueueTimer function 
 
 
 ## -description


PRMerger14:
@SpriteOvO : Thanks for your contribution! The author(s) have been notified to review your proposed change.

SpriteOvO:
I also just noticed that the argument types are in the correct order, only the names are in the wrong order.

Karl-Bridge-Microsoft:
Thanks for contacting us re. this issue. Unfortunately, I'm unable to reproduce the issue per the topic at https://docs.microsoft.com/en-us/windows/win32/api/threadpoollegacyapiset/nf-threadpoollegacyapiset-createtimerqueuetimer. Please see the following image, which shows everything in order.

image

SpriteOvO:
@Karl-Bridge-Microsoft Yes, it looks all right now, but it did have the issue before. Anyway, thanks for your checking.

PR #582, opened 2020-12-03
Update nf-cfapi-cfsetinsyncstate.md A parameter does not work as documented

LokiMidgard:
The InSyncUsn parameter does not work. The operation will fail using anything but NULL (including the correct USN).

After the method returns it will not write the new USN to the paramater in any case.

This is demonstrated by [this sample](https://github.com/LokiMidgard/CloudTest).

There was also a [stackoverflow](https://stackoverflow.com/questions/60816543/cfsetinsyncstate-does-not-behave-as-documented-cfapi) discussion, but this did not lead to any results. (It desscribed a workaround to get the USN from a file but not to run the operation conditionally).

### DIFF
--- a/sdk-api-src/content/cfapi/nf-cfapi-cfsetinsyncstate.md
+++ b/sdk-api-src/content/cfapi/nf-cfapi-cfsetinsyncstate.md
@@ -68,7 +68,7 @@ The in-sync state flags. See <a href="/windows/desktop/api/cfapi/ne-cfapi-cf_set
 
 ### -param InSyncUsn [in, out, optional]
 
-When specified, this instructs the platform to only perform in-sync setting if the file still has the same USN value as the one passed in. Passing a pointer to a USN value of 0 on input is the same as passing a NULL pointer.  On return, this is the final USN value after setting the in-sync state.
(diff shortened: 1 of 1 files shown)


PRMerger19:
@LokiMidgard : Thanks for your contribution! The author(s) have been notified to review your proposed change.

wangfu91:
@mikben 

I can confirm @LokiMidgard 's finding.

[CfSetInSyncState](https://docs.microsoft.com/en-us/windows/win32/api/cfapi/nf-cfapi-cfsetinsyncstate) API will fail if the ```USN``` parameter passed in is anything but NULL(or 0),
the error message is ```error code: 377, The file is not in sync with the cloud```

Even the correct USN obtained by ```DeviceIOControl``` with control code ```FSCTL_READ_FILE_USN_DATA``` will not work,

The ```CfSetInSyncState``` API call itself will generate a new USN record, so unless somehow I can predict the new USN value after the API called, and pass in this future USN value, there is no other way to make it work.

Please consider fixing the documentation or fix the implementation to avoid confusion and frustration.

MatchaMatch:

PR #586, opened 2020-12-08
Update the ShellExecuteExInfo flags to use a table

zadjii-msft:
Previously, these flags would be formatted as a header, and the docs site would make them absolutely massive. See [the docs page](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/ns-shellapi-shellexecuteinfoa#syntax)

image

They're bigger than the thing they're actually nested under.

I used an HTML table for the first list, the `fMask` flags, because multi-line content is WAY easier to include in an HTML table.

I used a markdown table for the `hInstApp` errors, because a Markdown one is generally easier to read in plaintext 😄

### DIFF
--- a/sdk-api-src/content/shellapi/ns-shellapi-shellexecuteinfoa.md
+++ b/sdk-api-src/content/shellapi/ns-shellapi-shellexecuteinfoa.md
@@ -69,49 +69,117 @@ Type: <b>ULONG</b>
 
 Flags that indicate the content and validity of the other structure members; a combination of the following values:
 
-#### SEE_MASK_DEFAULT (0x00000000)
+<table>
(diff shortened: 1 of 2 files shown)


PRMerger20:
@zadjii-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for your submission. I have incorporated your changes (with some modifications to address some formatting issues) along with some other general improvements to these articles in a PR that is being merged via our private repo for this doc set. The changes in that private PR should be published to the live docs site in the next day or two. Given this, I am closing this PR in the public repo. Thanks again for your help in improving these docs.

PR #592, opened 2020-12-12
CreateProcess: Make clear what happens when 0 is passed in the creation flags

nir9:
Problem: The CreateProcess functions have the creation flags parameter which offers a list of possible values but it doesn't specify what would happen when 0 is passed to this parameter.

Solution: Make clear that it is possible to pass 0 to this parameter and what it means.

### DIFF
--- a/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-createprocessa.md
+++ b/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-createprocessa.md
@@ -153,6 +153,9 @@ The flags that control the priority class and the creation of the process. For a
 This parameter also controls the new process's priority class, which is used to determine the scheduling priorities of the process's threads. For a list of values, see 
 <a href="/windows/desktop/api/processthreadsapi/nf-processthreadsapi-getpriorityclass">GetPriorityClass</a>. If none of the priority class flags is specified, the priority class defaults to <b>NORMAL_PRIORITY_CLASS</b> unless the priority class of the creating process is <b>IDLE_PRIORITY_CLASS</b> or <b>BELOW_NORMAL_PRIORITY_CLASS</b>. In this case, the child process receives the default priority class of the calling process.
 
+
+The dwCreationFlags parameter can be 0. In that case, the process creation will be default.
+
 ### -param lpEnvironment [in, optional]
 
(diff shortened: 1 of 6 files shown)


PRMerger17:
@nir9 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
Hi @nir9 - Can you confirm the "default"?

According to https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags, the default is "inheriting its parent's console (the default)".

nir9:
@Karl-Bridge-Microsoft Yah that is the default for the console applications but the inheritance has no meaning for windows applications.

I can add that and make it clear it happens with console apps

nir9:
Yah, that is one of the default behaviors but you have more stuff that happens when default is specified (as specified in the creation flags table).

I can make the change, but before that what do you think is more correct? specifying only the console inheritance (it is probably the most major thing that happens anyway) or maybe write it in a different way?

Karl-Bridge-Microsoft:
You should describe all scenarios. Not just the predominant one, which implies it's the only one.

PRMerger14:
@nir9 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

nir9:
@Karl-Bridge-Microsoft Added fixes that describe the meaning of passing 0 to the creation flags, also updated two more APIs I found that are similar.

PRMerger19:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger19:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger19:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger20:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger20:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #599, opened 2020-12-16
Update ns-winsvc-service_status.md

zodiacon:
The constant SERVICE_ACCEPT_USERMODEREBOOT does not seem to exist in the SDK header winsvc.h. Instead, this is defined SERVICE_ACCEPT_USER_LOGOFF with the same value (0x800).
Also, missing the control code that is sent to the service (SERVICE_CONTROL_USER_LOGOFF), which is commented out in the header.


PRMerger9:

PR #621, opened 2021-01-09
Add another scenario for when a DXGI_ERROR_INVALID_CALL is returned

siradam7th:
I added another scenario when CreateSwapChainForHwnd() returns an DXGI_ERROR_INVALID_CALL even when pDesc is not NULL, which is when one or more of the pDesc members are invalid.


PRMerger19:
@siradam7th : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #622, opened 2021-01-11
Update MSG_PEEK documentation from empirical observation

kyanha:
The behavior I'm documenting has been empirically seen on Windows 10 1909 x64. Arguably, it is a bug.

How should I submit a bug report on this behavior? I'd really prefer it be fixed to match the documentation as it was before this PR.

### DIFF
--- a/sdk-api-src/content/winsock2/nf-winsock2-recvfrom.md
+++ b/sdk-api-src/content/winsock2/nf-winsock2-recvfrom.md
@@ -307,7 +307,7 @@ The <i>flags</i> parameter can be used to influence the behavior of the function
 </tr>
 <tr>
 <td>MSG_PEEK</td>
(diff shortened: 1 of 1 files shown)


PRMerger19:
@kyanha : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
**UPDATE**: A contact in the product group tells me that the info this PR is suggesting to add is already present a few paragraphs above in the same topic. That leads me to believe that the behavior is not a bug, incidentally.

Hi, @kyanha! Looking at my list, it looks like [Microsoft Support](https://support.microsoft.com/contactus) is the resource for product bugs. That said, you might *try* one of the other resources listed below to report a bug if you prefer an alternative.

Here are all of the developer support resources that I'm aware of, for completeness.
- [Microsoft Q&A](https://docs.microsoft.com/answers/topics/windows-uwp.html) is the right place to lobby for a change or a new feature, to make a suggestion, or to ask a product question.
- For community assistance with development questions, there are our [forums](https://social.msdn.microsoft.com/Forums/newthread?category=windowsapps&forum=wpdevelop&prof=required), and [StackOverflow](https://stackoverflow.com/questions).
- The [Windows developer support](https://developer.microsoft.com/windows/support) site explains other support options.
- The [Visual Studio developer community](https://developercommunity.visualstudio.com) is available, and responsive. The former User Voice feature is now there on the **Suggest a feature** tab, if you want to lobby for a change, a feature, or to make a suggestion.
- For developer issues, or suspected product bugs, there's [Microsoft Support](https://support.microsoft.com/contactus); a support engineer will look at the repro and see whether the issue is a Microsoft product bug, or something else. If it's a bug with a Microsoft product then the ticket will be free of charge.

As for documenting bugs in the main content, that's something we avoid doing, and I believe that's because the content would then be incorrect as soon as the bug is fixed. Once they're notified of a bug, the product group can choose whether or not it's important enough to publish a knowledgebase article describing the issue.

So in light of that, I think it's best to go ahead and close this PR.

Thanks!
Steve

PR #628, opened 2021-01-12
format the table for better user experience

siradam7th:
I updated the table to make it easier to read its long content, Including updating the formatting for better readability when needed based on the context, also fixed an issue where the markdown code is outputted instead of a link.

**Reasons behind this change**:
* The second column `Description` in the table is way larger than the first one `Name`, making it a scrollable table, but because it is a very long table, the scrollbar is way down the page, not a user-friendly interaction for most screen sizes.
* This bothered me personally, but also was reported by another person on the DirectX discord mentioning the same issue, so i thought i'd open a pull request fixing all those issues at once.

**Things this edit fixes**:
* I fixed a markdown error that does not make the link show up, instead it showed the actual markdown code on the page.
* The overall user-experience is improved by using a one column table instead, the `Description` column was bigger than the `Name` column making it hard to read and follow, now it is way easier to browse and read.

**Reasons behind editing it this way**:
* I used `html` because markdown table syntax is not suitable due to the formatting of the page content.
* I had to use `<a href=""></a>` when needed because the markdown link syntax does not work inside `html` tags.

**Things to check before applying this edit**:
* [ ] Make sure that the C++ enum code at the top of the page under `Syntax` is outputted correctly, i could not check that because i have no idea how to build the docs, i guessed the previous code for `-enum-fields` and `-field` was responsible for it, so i kept at the top.
* [ ] Double-check the links and where they take (it was not possible to use the `[]()` syntax in some places because they are not parsed by markdown inside `html` tags. 

Feel free to comment further on this, thanks.

### DIFF
--- a/sdk-api-src/content/d3d12/ne-d3d12-d3d12_resource_flags.md
+++ b/sdk-api-src/content/d3d12/ne-d3d12-d3d12_resource_flags.md
@@ -46,109 +46,154 @@ api_name:
 ---
 
 # D3D12_RESOURCE_FLAGS enumeration
-
-
-## -description
-
-Specifies options for working with resources.
-
 ## -enum-fields
-
 ### -field D3D12_RESOURCE_FLAG_NONE
-
-No options are specified.
-
 ### -field D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET
-
-Allows a render target view to be created for the resource, as well as enables the resource to transition into the state of D3D12_RESOURCE_STATE_RENDER_TARGET. Some adapter architectures allocate extra memory for textures with this flag to reduce the effective bandwidth during common rendering. This characteristic may not be beneficial for textures that are never rendered to, nor is it available for textures compressed with BC formats. Applications should avoid setting this flag when rendering will never occur.
-
-
-The following restrictions and interactions apply:
-
-<ul>
-<li> Either the texture format must support render target capabilities at the current feature level. Or, when the format is a typeless format, a format within the same typeless group must support render target capabilities at the current feature level.</li>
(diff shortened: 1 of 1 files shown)


PRMerger9:
@siradam7th : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thank you! Well, let's go ahead and take this change, and see whether the tradeoffs make it better overall (if not I can revert it). The real solution here is to fix the generated tables, and we've had a bug on that for a while without progress as yet. As I say, while the solution here is not ideal, and is bound to look odd in some way, it's pretty bad right now, so if a strange-looking manual fix is the best we'll get for the foreseeable future, then it's worth taking.

Thanks!
Steve

siradam7th:
I understand that this is just a temporary workaround, I tried to pull every trick i could think of to come up with an 'acceptable' solution that works with all the content of the page uniformly.
To add to that, the way the markdown is skipped inside the `html` tags makes it hard to use `[]()`, so you'll have to fallback to the `<a>` html tag that needs an `href=""` instead of just pointing to the `.md` file which makes more sense when editing and keeping the links up-to-date in the long term.
 With that being said, I believe in the road to a real solution, in general, we could really benefit from being able to use markdown inside `html` tags selectively.

Thank you.

stevewhims:
Well, this was worth a try (I wanted to see how it'd look), but it didn't work. The topic ended up looking worse than before. I tried to revert the PR here, but the system wouldn't let me. So instead I've reverted the change via the internal PR https://cpubwin.visualstudio.com/win32/_git/sdk-api/pullrequest/17263 (which won't be visible publicly AFAIK).

As I mentioned above, the real solution here is to get the content build system to generate different output (output that renders more nicely), so I'll continue to pursue that with the content engineering team. There's no solution to that by changing the content of the topic, so continuing to try to fix it via PRs is a dead end. You'll note that there's no table in the original content. Only a list of headings with descriptions below each. Those headings/descriptions are turned into tables by the content build system. So that's the system we need to change. My recommendation here is for the build system to take the heading, make it bold, put a period after it, then render the description. That way there's no table (and no scrollbars) to go wrong.

Thanks again!
-Steve

PR #636, opened 2021-01-18
Update nf-sddl-convertsecuritydescriptortostringsecuritydescriptorw.md

tim-weis:
Fixed formatting.

And while there is no immediate way to provide feedback, let me abuse this issue to report, that the documentation for the `StringSecurityDescriptorLen` needs to be fixed as well (for both the -W as well as the -A API variant). The following issues need to be addressed:

* Remove references to `TCHAR`. Neither API variant uses the generic-text mappings.
* Make it clear what *"The size represents the size of the buffer in WCHARs, not the number of WCHARs in the string."* really means (is this just a convoluted way of saying "number of characters including the NUL terminator" or is there something more to it?).
* References to `WCHAR` need to be removed from the -A API variant (presumably).


PRMerger13:
@tim-weis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich (approved):

QuinnRadich:
Sorry for the delay on this.

Thanks for the feedback, Tim. I'm passing it on internally so it doesn't get abandoned here.

PR #648, opened 2021-01-28
Fix formatting in ITaskbarList3::SetProgressValue

krzysdz:
Because of indentation before the `<ul>` tag in Markdown, the tag was recognised as code in code block.

Screenshot of the code block containing indentation and an <ul> tag

Removing the indentation causes the tag to be correctly recognised as the beginning of an unordered list.\
Another (maybe better) way to fix it would be rewriting the HTML list using Markdown syntax.


PRMerger19:
@krzysdz : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #651, opened 2021-01-30
SwitchToThread() return value is retrospective

glenn-slayden:
It seems more correct to use the past tense when talking about the return value, since the "switching away" action has long-since occurred.


PRMerger9:

Karl-Bridge-Microsoft (approved):

PR #652, opened 2021-01-30
Gdi FloodFill() documentation refers to the last parameter as crFill …

ianhan:
…when it is "color."


PRMerger16:
@ianhan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ianhan:
It looks like ExtFloodFill() has the same issue, the parameter is named "color" but referred to in the documentation as "crColor."

hkhmsft:
Thanks for the contribution!

PR #660, opened 2021-02-05
Update ns-commdlg-openfilenamea.md

ratchetfreak:
codify the example filter strings so people stop being confused by stray italic text and missing asterisks.

I just came from helping someone being confused by this and only barely caught the italics hinting at asterisks that needed to be in the example string. 

So I added code tags to match the other example string where the asterisks are properly rendered and escaped the stray one to avoid future issues.


PRMerger8:
@ratchetfreak : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys (approved):

PR #666, opened 2021-02-11
Update nf-winuser-setfocus.md

NikoAri:
Documentation should specify what causes different errors. For example, during debugging of large app, I got ERROR_INVALID_PARAMETER, it it's really not obvious why.

Adding comment:
Extended error ERROR_INVALID_PARAMETER (0x57) means that window is in disabled state.


PRMerger14:

jwmsft (approved):

PR #679, opened 2021-02-20
Docs page confuses GetLastError with return value

jnm2:
The docs page is wrong when it lists ERROR_INVALID_PARAMETER as a "return value" and fails to mention GetLastError: https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-createdibsection#return-value

You can verify this by passing an invalid value to the `usage` parameter, such as 12345. CreateDIBSection then returns `0` and GetLastError returns ERROR_INVALID_PARAMETER.

Hans Passant (a reliable source in this area) says the same thing here: https://stackoverflow.com/questions/7932153/windows-gdi-context-createdibsection#comment9693188_7932153

### DIFF
--- a/sdk-api-src/content/wingdi/nf-wingdi-createdibsection.md
+++ b/sdk-api-src/content/wingdi/nf-wingdi-createdibsection.md
@@ -120,13 +120,13 @@ The offset from the beginning of the file-mapping object referenced by <i>hSecti
 
 If the function succeeds, the return value is a handle to the newly created DIB, and *<i>ppvBits</i> points to the bitmap bit values.
 
-If the function fails, the return value is <b>NULL</b>, and *<i>ppvBits</i> is <b>NULL</b>.
+If the function fails, the return value is <b>NULL</b>, and *<i>ppvBits</i> is <b>NULL</b>. To get extended error information, call <a href="/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a>.
 
-This function can return the following value.
+<a href="/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a> can return the following value:
 
 <table>
 <tr>
-<th>Return code</th>
+<th>Error code</th>
 <th>Description</th>
 </tr>
 <tr>


PRMerger19:
@jnm2 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

AraHaan:
I feel like the api function to this should be patched to return FALSE when it fails just like all the other winapi functions as that is what every developer would expect (to check error only when a windows api call returns FALSE so that way we can then check GetLastError() since calling GetLastError() unconditionally could give an invalid or incorrect code (especially if it is called after an api function, but unknowingly the user expects it to internally call SetLastError, however it never is called and it results in the last error code set from some other api function call to be returned).

jnm2:
@AraHaan It already returns zero when it fails, and I think its behavior can't be patched anyway without breaking decades of applications. This PR only tracks aligning the docs with reality of the existing behavior.

jayv-ops:
Guys,
I am not a developer or tech guy.
I have been hacked for over a yr now they are in every single bit of equipment i own, my house looks like a computer grave yard where they change the passwords and everything else i have lost most of my email accounts this one works intermittently, I have even changed my phone number and provider 3 time in the last yr. They have even got into my cloud accounts so I'm completely buggered.  
They originally got in through my internet hub i have been told.
To be honest I'm completely depressed with this entire thing this all started just as my father was passing away so just what i didn't need. 
I have reached out to so many people including the police for help yet here i am still in the same shit and keep being told due covid theres nothing they can currently do.
I have found multiple accounts and sites that i have never even heard of appear on my history in my name so i usually go on and try and delete anything i can to try and stop them. Apparently i now have a facebook and twitter accounts which i have never had or wanted.  So i apologies if i have upset anyone but i don't know what else to do.
James

jayv-ops:
Sorry one more thing i have never agreed or given my consent to any of this stuff either 
James

GrantMeStrength:
Thanks for submitting this correction, and for the link to the Stackoverflow issue. I am awaiting confirmation. -john

riverar (approved):

GrantMeStrength:
Thanks!

PR #681, opened 2021-02-24
Removed trailing commas

AlexWilkinsonnn:
Trailing commas are not valid JSON


GrantMeStrength:
Good catch, thank you!

-John

PR #691, opened 2021-03-02
Update nf-wlanapi-wlansetprofileeapxmluserdata.md

soshogoh:
I have found that  TtlsCfg always use 64 bit buffer to convert the user BLOB.
This API should mentioned that for EAP-TTLS, the application builds should be the same as OS Platform.
There is also a forum that discussed this problem:
WlanSetProfileEapXmlUserData for EAP-TTLS with inner method EAP-MSCHAPv2(not MSCHAPv2)
https://social.msdn.microsoft.com/Forums/en-US/f013931c-e9cc-4c14-bc62-363219c9dd69/wlansetprofileeapxmluserdata-for-eapttls-with-inner-method-eapmschapv2not-mschapv2

Doc fix suggestion:
------
The WlanSetProfileEapXmlUserData may cause wireless conenction fail when use EAP-TTLS if the API is called from 32-bit builds application but the OS Platform is 64-Bit. Your application should be built by the same CPU architecture of the OS Platform.
------

### DIFF
--- a/sdk-api-src/content/wlanapi/nf-wlanapi-wlansetprofileeapxmluserdata.md
+++ b/sdk-api-src/content/wlanapi/nf-wlanapi-wlansetprofileeapxmluserdata.md
@@ -45,18 +45,15 @@ api_name:
  - WlanSetProfileEapXmlUserData
 ---
 
-# WlanSetProfileEapXmlUserData function
-
-
 ## -description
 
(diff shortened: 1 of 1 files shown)


PRMerger13:
@soshogoh : Thanks for your contribution! The author(s) have been notified to review your proposed change.

soshogoh:
Since the Merging is blocked for me, Please help to merge this request.

stevewhims (approved):

PRMerger17:

PR #706, opened 2021-03-12
LCMapStringW doc needs clarifying

code-creator:
The description of return values is wrong when input character count does not include terminating zero, and doc should mention in this case no terminating zero is written to destination string.

See test app below (compile as Unicode main.cpp):

#include \<iostream\>
#include \<windows.h\>

int wmain() {

	int reqlen = LCMapStringW(LOCALE_INVARIANT, LCMAP_LOWERCASE | LCMAP_LINGUISTIC_CASING, L"Hi", 2, nullptr, 0);
	std::wcout << "Returns required len without space for terminating zero:" << reqlen << std::endl;

	wchar_t dst[3];
	memset(dst, 0xFF, sizeof(dst));
	LCMapStringW(LOCALE_INVARIANT, LCMAP_LOWERCASE | LCMAP_LINGUISTIC_CASING, L"Hi", 2, dst, 3);
	std::wcout << "Even if space for terminating zero is in dst, does not write it:" << std::hex << (unsigned int)dst[2] << std::endl;

	return 0;
}
`

### DIFF
--- a/sdk-api-src/content/winnls/nf-winnls-lcmapstringw.md
+++ b/sdk-api-src/content/winnls/nf-winnls-lcmapstringw.md
@@ -113,6 +113,7 @@ The application cannot set this parameter to 0.
 ### -param lpDestStr [out, optional]
 
 Pointer to a buffer in which this function retrieves the mapped string or a sort key. When the application uses this function to generate a sort key, the destination string can contain an odd number of bytes. The LCMAP_BYTEREV flag only reverses an even number of bytes. The last byte (odd-positioned) in the sort key is not reversed.
+Note when the <i>cchSrc</i> parameter does not include space for the terminating null character, no terminating null character is written to <i>lpDestStr</i>, even if the buffer size given by the parameter <i>cchDest</i> would allow it.
 
 <div class="alert"><b>Note</b>  The destination string can be the same as the source string only if LCMAP_UPPERCASE or LCMAP_LOWERCASE is set. Otherwise, the strings cannot be the same. If they are, the function fails.</div>
 <div> </div>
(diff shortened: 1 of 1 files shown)


PRMerger8:
@code-creator : Thanks for your contribution! The author(s) have been notified to review your proposed change.

code-creator:
In first line of test app, the line #include \<iostream\> got scrambled.
Updated now.

If you run the test app, it prints:
_Returns required len without space for terminating zero:2
Even if space for terminating zero is in dst, does not write it:ffff_

According to the original documentation of **LCMapStringW** , the expected value would be three for the first line.
If the terminating zero is written when dst has enough space was not specified, so the expected value of second line 
would be not known.

code-creator:
Comparing the documentation of return values of LCMapStringW with LCMapStringEx, there the case of getting the buffer size is correct regarding terminating zero in input string (bold text is missing in LCMapStringW documentation):

"Returns the number of characters or bytes in the translated string or sort key, including a terminating null character, if successful. If the function succeeds and the value of cchDest is 0, the return value is the size of the buffer required to hold the translated string or sort key, including a terminating null character **if the input was null terminated**."

But there also is stated the return value includes terminating null character for string mapping, which is by my tests incorrect if called:
`int r= LCMapStringEx(LOCALE_NAME_INVARIANT, LCMAP_LOWERCASE | LCMAP_LINGUISTIC_CASING, L"Hi", 2, dst, 3, nullptr, nullptr, 0L);`
it returns two, and also does not write a terminating zero.

ShawnSteele:
I don’t like the wording around the return not including null.  The return will also include the null terminator if the input length was -1 (eg: null terminated).  It’s only when the caller explicitly requested a particular subset of the string that we don’t include a null (unless the caller included it in their count.)

ShawnSteele:
Not relevant to this particular change, but this section:

"The destination string can be the same as the source string only if LCMAP_UPPERCASE or LCMAP_LOWERCASE is set. Otherwise, the strings cannot be the same. If they are, the function fails."

Should probably qualify that this is discouraged, something like:

"Though the API allows the destination string to be the same buffer as the source string when LCMAP_UPPERCASE or LCMAP_LOWERCASE is set this behavior is strongly discouraged as some conditions may cause the returned cased string to be a different length."

Karl-Bridge-Microsoft:
@code-creator - based on feedback and the need to make the same updates to the nf-winnls-lcmapstringa.md file, I created an internal PR to handle this. I'll close this PR, but please open another on the updated files if necessary.

PR #716, opened 2021-03-22
Patch 2

code-creator:
I have tried to improve documentation:

-    warn terminating null not written when used for string mapping and source string size does not include terminating null
-    remove comment of sort key being null terminated, as documentation already states it can contain embedded nulls at any position
-    describe how to compare sort keys
-    better describe the return values
-    better emphasis when return values are character count/byte counts

I'm not English speaker so please correct any mistakes I made.
The documentation for this function is really complicated, and one could consider to completely separate the documentation for using it for string mapping vs. using it for sort key generation ;-)

### DIFF
--- a/sdk-api-src/content/winnls/nf-winnls-lcmapstringex.md
+++ b/sdk-api-src/content/winnls/nf-winnls-lcmapstringex.md
@@ -240,7 +240,7 @@ The application cannot set this parameter to 0.
 
 ### -param lpDestStr [out, optional]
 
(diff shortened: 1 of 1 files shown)


PRMerger18:
@code-creator : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
@code-creator - thanks for the suggestions. I've updated this topic and others in a private PR. I'm going to close this PR, but please open a new one on the revised topics if necessary.

PR #721, opened 2021-03-27
ImpersonateAnonymousToken expects THREAD_IMPERSONATE right before sta…

GeoB99:
…rting the impersonation

As of now, the documentation of `ImpersonateAnonymousToken` service is vague at best. It is mentioned a valid thread handle has to be filled to the caller but no kind of access right mask is specified that the thread needs to have in order to begin the impersonation procedure.

As a matter of fact, this public Win32 service bases upon the `NtImpersonateAnonymousToken` NT system call in order to impersonate the system's anonymous logon token. That system call invokes the kernel mode service from the Object Manager, the `ObReferenceObjectByHandle`, with desired access mask of `100h` value which is equivalent to `0x0100` and the documentation [here](https://docs.microsoft.com/en-us/windows/win32/procthread/thread-security-and-access-rights) demonstrates that. 

If the thread handle object has such access then the system call happily grants impersonation to the calling thread, otherwise it outright fails with a NTSTATUS code of `STATUS_ACCESS_DENIED` (further translated and mapped as `ACCESS_DENIED` in Win32 world). With that being said, it doesn't hurt for the documentation to be accordingly updated with the remarks based off of these points.

### DIFF
--- a/sdk-api-src/content/securitybaseapi/nf-securitybaseapi-impersonateanonymoustoken.md
+++ b/sdk-api-src/content/securitybaseapi/nf-securitybaseapi-impersonateanonymoustoken.md
@@ -6,7 +6,7 @@ helpviewer_keywords: ["ImpersonateAnonymousToken","ImpersonateAnonymousToken fun
 old-location: security\impersonateanonymoustoken.htm
 tech.root: security
 ms.assetid: 98d1072e-f569-4c8c-9254-fa558054c7ec
-ms.date: 12/05/2018
+ms.date: 03/28/2021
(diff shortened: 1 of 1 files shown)


PRMerger9:
@GeoB99 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PRMerger12:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #732, opened 2021-04-10
Correct number of characters vs. bytes

vid512:
According to my test, this function writes at most 1024 wide characters, not bytes. That was probably just a copy-paste from wsprintfA.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-wsprintfw.md
+++ b/sdk-api-src/content/winuser/nf-winuser-wsprintfw.md
@@ -61,7 +61,7 @@ Writes formatted data to the specified buffer. Any arguments are converted and c
 
 Type: <b>LPTSTR</b>
 
-The buffer that is to receive the formatted output. The maximum size of the buffer is 1,024 bytes.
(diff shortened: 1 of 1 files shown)


PRMerger6:
@vid512 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger8:
@vid512 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

justanotheranonymoususer:
From my tests, 1024 characters are written and the null terminator is then appended, resulting in 2050 bytes being written.

I opened PRs about it a while ago:
https://github.com/MicrosoftDocs/sdk-api/pulls?q=is%3Aopen+is%3Apr+author%3Ajustanotheranonymoususer+nf

vid512:
Yes, you are correct, I have overlooked the terminating zero. It would be better to add single explicit sentence about what happens in this case: only first 1024 characters + terminating zero are written, value 1024 is returned... This is not immediately obvious just from description "maximum buffer size" given in bytes. Anyway, your PR is more complete, so I am closing this.

PR #734, opened 2021-04-13
Update nf-winuser-settimer.md

NikoAri:
Problem:
By default, Windows sets up timer procs in a __try/__except statement, where it catches any SEH errors, including Access Violation, and swallows them. Note, that if users use C++, this problem is further complicated because __try/__except doesn't do any stack unwinding, and thus creates an unknown and corrupt app state. Some of the effects of skipping stack unwind are objects leaks, mutexes not released, pointers, files, not cleaned up, etc. Long story short, at this point app is in unknown state. After SEH is swallowed, Windows Apps go back to message processing as if nothing has happened, but of course App is still in an unknown, potentially corrupt state. Problems that are associated with such corrupt states are extremely difficult to investigate.

To avoid the above described problem, Microsoft recommends setting UOI_TIMERPROC_EXCEPTION_SUPPRESSION to false:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setuserobjectinformationw
However, that recommendation is not linked to SetTimer Api in any way, thus is hard to discover.

Solution:
Microsoft SetTimer documentation should promote its recommendation to set Windows UOI_TIMERPROC_EXCEPTION_SUPPRESSION flag to false before using SetTimer Api.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-settimer.md
+++ b/sdk-api-src/content/winuser/nf-winuser-settimer.md
@@ -110,8 +110,8 @@ The <i>wParam</i> parameter of the <a href="/windows/desktop/winmsg/wm-timer">WM
 The timer identifier, <i>nIDEvent</i>, is specific to the associated window. Another window can have its own timer which has the same identifier as a timer owned by another window. The timers are distinct. 
 
 <b>SetTimer</b> can reuse timer IDs in the case where <i>hWnd</i> is <b>NULL</b>. 
-			
-
+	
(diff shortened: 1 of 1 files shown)


PRMerger7:

NikoAri:
Similar change should be made for **SetCoalescableTimer** Api:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcoalescabletimer

jwmsft (approved):

PRMerger13:
@jwmsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #741, opened 2021-04-18
Update nf-winbase-lstrcmpw.md

roman380:


PRMerger13:
@roman380 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich:
I do not believe the requested change is accurate. https://docs.microsoft.com/en-us/windows/win32/intl/windows-data-types-for-strings covers the definition of the string types in question, but my research indicates that the TSTR format is accurate (and is functionally identical to the WSTR format in unicode-supported scenarios)

PR #746, opened 2021-04-21
Update API doc for SystemParametersInfoA/SystemParemetersInfoW on SPI_SETTHREADLOCALINPUTSETTINGS parameter

jimmyzhush:
When we call APIs SystemParametersInfoA/SystemParemetersInfoW to set active input settings using SPI_SETTHREADLOCALINPUTSETTINGS action, the parameter pvParam accepts a BOOL variable casted by PVOID instead of a pointer to a BOOL variable. If we use a pointer to BOOL variable here, the APIs always treat it as Local (per-thread, TRUE) because a pointer to BOOL variable cannot be zero.


PRMerger13:
@jimmyzhush : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #752, opened 2021-04-26
Message fixes: file-not-found 2 [BULK UPDATE]

TimShererWithAquent:
Anybody notified by GitHub, this is for [1808303](https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1808303). We're addressing various suggestions around accessibility and SEO. Your feedback is welcome, but not required for approval of these changes.

Addresses **file-not-found**.

This is a bulk update.


PRMerger12:
@TimShererWithAquent : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (changes requested):
The double-dots (navigate up a folder) won't work. The single dots that are already there would be correct. But the live topics (at least, I spot-checked just the first) already work.

I also saw one case where you were using file paths, but omitted the .md.

TimShererWithAquent:
Hi @stevewhims - This has been a bit of a mess - There are broken links on the live site, but only the ones that we're fixing from the validation report. We'll close this PR soon and get the validation resolved. 

Sorry for the inconvenience -

PR #753, opened 2021-04-27
TIMERPROC fix unnamed parameters

fdwr:
The proper parameter names are missing. Technically the function prototype in windows.h *is* missing these names (`typedef VOID (CALLBACK* TIMERPROC)(HWND, UINT, UINT_PTR, DWORD);`), but you can tell from `SetTimer` what they're supposed to be (https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer). Note the naming distinction though between `dwTime` vs `uElapse`, since `dwTime` here is the current time since system startup, as opposed to SetTimer `uElapse` which is the timer interval.


PRMerger20:

MatchaMatch:

PR #777, opened 2021-05-21
Update ns-wingdi-displayconfig_path_info.md

mr-intj:
Request for clarification or reference link.

Also, there used to be a *Submit Feedback (about) This Page* button at the bottom of these pages. I'm glad that I can now make direct edits and submit an MR when I have a specific fix to propose. However, it's not clear to me what the process is now for requesting *additional information*. Inserting my comments inline into the doc and submitting an MR feels like misuse -- please let me know if that's the case, and what the intended process is. Thanks.


PRMerger14:

hkhmsft:
forwarding feedback.

PR #778, opened 2021-05-21
Update ne-wingdi-displayconfig_device_info_type.md

mr-intj:
Requests to fill out documentation for three enumerands and add pages for two struct types (see comments inline).

Also, there used to be a button at the bottom of many docs.microsoft.com pages to generate a GitHub issue with links back to the docs page. That seems to be gone now, and it's not clear what your preferred procedure is for us to request doc changes that we can't entirely edit & submit on our own. Please let me know, and I'll use that going forward vs. creating an MR.


PRMerger9:

hkhmsft:
Hi @mr-intj , we currently only have issues turned on in our conceptual repos. I will forward this to our product team and they will add the documentation. Many thanks!

PR #783, opened 2021-05-26
Remove "api requires packaging" errors

jonwis:
Metadata resolution APIs work for packaged and unpackaged applications.  Some metadata lookup uses the package graph, but "do the right thing" if the process has no package graph.


PRMerger16:
@jonwis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #786, opened 2021-05-28
Clarify that packed mips aren't a single tile

AdamJMiles:
Clarifying that the packed mips on a texture aren't necessarily packed into a single tile, but rather multiple tiles.


PRMerger12:
@AdamJMiles : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #795, opened 2021-06-04
Fix incorrectly escaped backslash

JensMertelmeyer:
Text was supposed to contain a backslash character in quotes. `'\'` was wrong, `'\\'` should be correct.


PRMerger17:
@JensMertelmeyer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #801, opened 2021-06-09
Add the return value in case of success

alessandro-gentilini:
I would like to submit an issue but here at https://github.com/MicrosoftDocs/sdk-api I do not see the "Issues" tab
image

The page
https://docs.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-isequentialstream-write has an incomplete *Return value* section. It states

> This method can return one of these values.

and does not list any values.
After a rapid browsing it seems to me that the problem is also at any method of the interface `IStream`, `IStorage`...

### DIFF
--- a/sdk-api-src/content/objidl/nf-objidl-isequentialstream-write.md
+++ b/sdk-api-src/content/objidl/nf-objidl-isequentialstream-write.md
@@ -68,7 +68,8 @@ A pointer to a <b>ULONG</b> variable where this method writes the actual number 
 
 ## -returns
 
-This method can return one of these values.
+This method can return one of these values. 
+S_OK in case of success.
 
 ## -remarks
 


PRMerger6:
@alessandro-gentilini : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for reporting the missing return value info from these interfaces. We are investigating and trying to determine whether there is some old content we can bring back into the published docs.

mcleanbyron:
I recovered the missing return values for ISequentialStream.Write and 41 other APIs in this same section of the docs. The missing return value information should be published on the live docs site in the next day or two. Given this, I am closing this pull request. 

In the meantime, here is the complete table of return value information for ISequentialStream.Write: 

| Return code | Description |
|----------------|---------------|
|S_OK | The data was successfully written to the stream object.|
|E_PENDING | Asynchronous Storage only: Part or all of the data to be written is currently unavailable.|
|STG_E_MEDIUMFULL | The write operation failed because there is no space left on the storage device.|
|STG_E_ACCESSDENIED | The caller does not have the required  permissions for writing to this stream object.|
|STG_E_CANTSAVE | Data cannot be written for reasons other than improper access or insufficient space.|
|STG_E_INVALIDPOINTER | One of the pointer values is not valid. The *pv* parameter must contain a valid pointer even if *cb* is zero.|
|STG_E_REVERTED | The object has been invalidated by a revert operation above it in the transaction tree.|
|STG_E_WRITEFAULT | The write operation failed due to a disk error. This value is also returned when this method attempts to write to a stream that was opened in simple mode (using the STGM_SIMPLE flag).|

PR #805, opened 2021-06-12
Fix `InternetSetStatusCallback` return value

justanotheranonymoususer:
I'm sorry for misusing the pull requests mechanism, but it's an obvious error and I don't know how to fix it.

The `InternetSetStatusCallback` function is marked as returning `void` [in the docs](https://docs.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-internetsetstatuscallback), but in reality it returns `INTERNET_STATUS_CALLBACK`.


PRMerger12:

mcleanbyron:
Thanks for reporting this to us. Although the syntax section for this API is incorrectly list the return value as 'void', the description section is accurate. I took a look at the source for this, and it appears that the docs build pipeline that parses the header and generates the code in the syntax section isn't handling the `INTERNETAPI_(type)` macro in the header correctly. I have opened a bug against our internal docs build system to see if we can get this fixed. In the meantime, I am closing this pull request because we can't edit this information manually in the article. Not that this also affects other functions that are defined using the same macro in the source, such as [InternetOpenA](https://docs.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-internetopena).

PR #813, opened 2021-06-22
Use NetSh not NetShell to refer to netsh.exe

dthaler:
netshell.dll is unrelated to netsh.exe, so netsh.exe should never be referred to as "netshell".


PRMerger18:
@dthaler : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #816, opened 2021-06-23
Update ne-winnt-token_type.md

jeanga:
Adding = 1 to the definition of  TokenPrimary helps understand that the first value is __not__ 0 (default) but 1.
As is, the definition does not represent the reality of this enum in winnt.h:
```cpp
typedef enum _TOKEN_TYPE {
    TokenPrimary = 1,
    TokenImpersonation
    } TOKEN_TYPE;
typedef TOKEN_TYPE *PTOKEN_TYPE;
```


PRMerger20:
@jeanga : Thanks for your contribution! The author(s) have been notified to review your proposed change.

DJm00n:
AFAIK ‘### -field’ tags are special and you not allowed to add anything there. Just put that value somewhere in description.

PR #818, opened 2021-06-26
`SERVICE_TABLE_ENTRYA`: `lpServiceName` should be of type `LPCSTR`

justanotheranonymoususer:
A bug report. The docs are in sync with the headers, but the headers are incorrect. I don't know where to report it, so reporting it here. Hopefully one of the Microsoft guys here can forward it to the right location. Thanks.


PRMerger8:

Robo210:
While you are right that this (and every other string in the header) likely should have been const, changing them now would be a breaking change for existing code.

justanotheranonymoususer:
Why? A non-const string is implicitly casted to const string, no?

GrantMeStrength:
I will hold off changing them for now, thanks.

PR #829, opened 2021-07-06
Fixes formatting

tim-weis:


mcleanbyron (approved):

tim-weis:
7 days in between proposing a change and having it merged. The turn-around times have gone down considerably. Thanks a lot, this is much appreciated!

One question, though, if you don't mind: Is there a public document describing where this documentation came from, how it was translated, and most importantly, what the preferred formatting is (Markdown vs. HTML), and why?

mcleanbyron:
We don't have something like this specifically for the Win32 API docs (the subject of this particular repo), but the [Microsoft Docs contributor guide](https://docs.microsoft.com/en-us/contribute/) has some general info that applies generally to all Microsoft technical docs.

PR #852, opened 2021-07-22
Update VertexFormat requirements for DXR Tier 1.1

maamert:
DXR Tier 1.1 extends support for D3D12_RAYTRACING_GEOMETRY_TRIANGLES_DESC::VertexFormat to more formats. This change updates the documentation to reflect this. All wording was copy-pasted directly from the DXR spec.


PRMerger7:
@maamert : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Ideally we would use markdown, but since there's already a ton of HTML in our markdown files, I won't block approval and completion. :) Thank you!

PR #855, opened 2021-07-24
Separate predefined key list with commas

zodiacon:
The predefined key list appeared glued together. I added commas so these are properly separated.


PRMerger8:

mcleanbyron (approved):

PRMerger6:

mcleanbyron:
Thank you for your edits. In this case, I think a list will be more easily scanned by readers, and it looks like that was the intent of the original content. I added Markdown-formatted bullets for the list so it should render correctly.

PR #858, opened 2021-07-28
Update nf-shellapi-shellexecutea.md and nf-shellapi-shellexecutew.md

CookiePLMonster:
The recommendation to cast `HINSTANCE` to `int` confuses people since it generates a truncation warning, as evidenced e.g. by StackOverflow topics like [this one](https://stackoverflow.com/questions/67098202/avoid-msvc-warning-when-casting-shellexecute-return-hinstance-to-int).

Referring to `INT_PTR` instead should clear up the confusion, remove the need for oddly-looking double casts and/or warning suppression, and [still aligns with Raymond Chen's recommendations](https://devblogs.microsoft.com/oldnewthing/20061108-05/?p=29083):
> Or you could cast the result to an INT_PTR and compare the result against 32. That’s fine, too.

### DIFF
--- a/sdk-api-src/content/shellapi/nf-shellapi-shellexecutea.md
+++ b/sdk-api-src/content/shellapi/nf-shellapi-shellexecutea.md
@@ -142,7 +142,7 @@ The flags that specify how an application is to be displayed when it is opened. 
 
 Type: <b>HINSTANCE</b>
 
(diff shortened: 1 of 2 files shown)


PRMerger6:
@CookiePLMonster : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #861, opened 2021-08-03
Fix documentation of WinHTTP Ex API return codes

Techwolfy:
The new WinHTTP Ex APIs return a DWORD containing a Win32 error code, not a BOOLEAN success value.


PRMerger8:
@Techwolfy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #866, opened 2021-08-10
ADD1 did not support INT32/UINT32

fdwr:
This was copy pasta :/. We can see in the actual codebase that the HLSL shaders only support float16/float32. (if we add these data types in the future, then a future feature level >=4.1 will support them) OS Bug 33358314


PRMerger9:

stevewhims (approved):

PR #867, opened 2021-08-10
Update nf-d3d11_3-id3d11query1-getdesc1.md

rpavlik:
Point to the base C desc type, not the C++ wrapper/helper.

It's nice to know about the wrapper, but the docs for the wrappers don't have links to the member types. This is the first struct link I've followed in the d3d11 docs that goes to the C++ wrapper, so I guessed it was an accident.


PRMerger13:
@rpavlik : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #880, opened 2021-08-27
Update nf-d3d11-id3d11device-checkmultisamplequalitylevels.md

Const-me:
I hope this version is correct.
At least that’s the behavior I’m observing on my PC with Ryzen 7 5700G CPU, GeForce 1080Ti GPU, and Windows 10 20H2.

### DIFF
--- a/sdk-api-src/content/d3d11/nf-d3d11-id3d11device-checkmultisamplequalitylevels.md
+++ b/sdk-api-src/content/d3d11/nf-d3d11-id3d11device-checkmultisamplequalitylevels.md
@@ -78,7 +78,7 @@ This method returns one of the following <a href="/windows/win32/direct3d11/d3d1
 
 ## -remarks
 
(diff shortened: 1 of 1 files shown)


PRMerger6:
@Const-me : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Since there's uncertainty, I'll get a technical review. Please stand by.

-Steve

clandrew:
Really good catch!

CheckMultisampleQualityLevels's returned HRESULT is based on the app passing in good parameters understood by the runtime(e.g., not a sample count of 0, or some garbage DXGI_FORMAT). The returned HRESULT is not really based on anything the driver says. 

So yeah, it is possible to have CheckMultisampleQualityLevels return S_OK since your parameters were all good, while yielding a numQualityLevels of 0 since that's what the driver reported.

Good for docs to indicate this. Thank you!

Const-me:
@stevewhims OK, but I’m pretty sure their response is going to be “that’s up to GPU driver to decide”.

For instance, in this very article Microsoft says “FEATURE_LEVEL_11_0 devices are required to support 8x MSAA”, however in reality, Intel Haswell GPU, which [on paper](https://en.wikipedia.org/wiki/Intel_Graphics_Technology#Capabilities_(GPU_hardware)) supports D3D 11.1, [only supports 4x MSAA](https://community.intel.com/t5/Graphics/With-8x-MSAA-read-from-multisample-texture-in-D3D-11-silently/td-p/1100648).

stevewhims (approved):

PR #885, opened 2021-09-03
Fixed grammar, and added clarity

badasahog:


PRMerger17:
@badasahog : Thanks for your contribution! The author(s) have been notified to review your proposed change.

badasahog:
I don't know how to update a pull request, but I found another typo:

The debug layer issues an error if it detects that any resource referenced by the command lists, including queries, has been destroyed.

should be changed to 

The debug layer issues an error if it detects that any resource referenced by the command lists, including queries, have been destroyed.

stevewhims (approved):

PR #901, opened 2021-09-15
Add color format to sampler desc BorderColor field

hannes-harnisch:
The BorderColor field doesn't tell you what color format the float array is being interpreted as.


PRMerger18:
@hannes-harnisch : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #925, opened 2021-10-12
Reference WSAOVERLAPPED rather than _WSAOVERLAPPED

EricScottSalem:
_WSAOVERLAPPED is the WIN16 version of WSAOVERLAPPED.  Internal and InternalHigh are DWORDs rather than ULONG_PTR.  As a result, the documentation page contradicts itself (struct definition vs struc-fields description).
That being said, I have never used this documentation system and do not know what the policy is for conditionally defined structures, nor how to update the documentation to properly reflect the WIN32 WSAOVERLAPPED structure.  Assistance is appreciated.

### DIFF
--- a/sdk-api-src/content/winsock2/ns-winsock2-wsaoverlapped.md
+++ b/sdk-api-src/content/winsock2/ns-winsock2-wsaoverlapped.md
@@ -1,5 +1,5 @@
 ---
-UID: NS:winsock2._WSAOVERLAPPED
+UID: NS:winsock2.WSAOVERLAPPED
 title: WSAOVERLAPPED (winsock2.h)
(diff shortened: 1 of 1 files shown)


PRMerger14:
@EricScottSalem : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, @EricScottSalem, but we're ok with this topic as-is. We don't need this change.

The `UID:` field is 100% an internal implementation detail belonging to the content build/publishing system. There's no reason to expect the value of that field to correspond to an actual public API name.

In fact if we were to change this value like this, then this topic would disappear from the live site until we changed the UID field back.

I'll go ahead and close the PR, since we don't want to take this change.

Thanks!
Steve

PR #926, opened 2021-10-12
Add description of structured buffer behavior of NumSamples

FreakingBarbarians:
Spoke with @jenatali on the DX12 Discord. The compiler repurposes NumSamples to store the stride of the structured buffer struct.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-setthreadcursorcreationscaling.md
+++ b/sdk-api-src/content/winuser/nf-winuser-setthreadcursorcreationscaling.md
@@ -31,7 +31,7 @@ topic_type:
 api_type:
  - 
 api_location:
- - winuser.h
+ - user32.dll
 api_name:
  - SetThreadCursorCreationScaling
 f1_keywords:


PRMerger8:
@FreakingBarbarians : Thanks for your contribution! The author(s) have been notified to review your proposed change.

sdk-api-src/content/d3d12shader/ns-d3d12shader-d3d12_shader_input_bind_desc.md
```diff
@@ -88,6 +88,7 @@ A <a href="/windows/desktop/api/d3dcommon/ne-d3dcommon-d3d_srv_dimension">D3D_SR
 
 The number of samples for a multisampled texture; when a texture isn't multisampled, the value is set to -1 (0xFFFFFFFF).
             This is zero if the shader resource is not a recognized texture.
+            If the resource is a structured buffer the field returns the stride of the struct.
```
Triang3l: The comma between "buffer" and "the" is missing (in the wrong place in the newer version).

Triang3l: "Returns" doesn't sound like the correct term for a structure field, as it's the terminology used for functions (which return flow control as well as optionally a value to the caller). The two lines above reference the field as a container ("the value is set") or as a value directly ("this is zero"), I think "is set to" is applicable here too.

Triang3l: I don't think there's need to truncate words in documentation, especially to shorten one from 9 to 6 characters. In the context of structured buffers, the word commonly used in Microsoft documentation is "structure". `struct` is an HLSL keyword for a language concept, and as far as I'm aware, you can also declare a `StructuredBuffer<float4>` or other kind of a structured buffer with elements of numeric types, which will not involve an HLSL `struct`. The `D3D12_BUFFER_SRV` documentation says "size of each element in the buffer structure", though this wording is pretty confusing too (is the structure an array element, or actually the array itself?)

Triang3l: Is it in bytes?

Triang3l (changes requested):

PRMerger10:
@FreakingBarbarians : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #932, opened 2021-10-13
DirectXMath matrix view function updates

walbourn:
The 'nearZ' and 'farZ' values can be reversed for "reverse z buffer" rendering setups.


PRMerger12:
@walbourn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
I'm not sure about removing the spaces from the YAML metadata. It's probably still valid, but I can't be certain. In future I'd recommend steering clear of making any changes to metadata unless they're absolutely necessary. But I'll optimistically approve and merge this since as I say it's probably ok.

Other thoughts for future: we like to contract (can't rather than cannot), and prefer markdown to HTML.

walbourn:
Atom removed the trailing whitespace. Do you need a particular setting for editing these?

PR #942, opened 2021-10-25
Task 29854083: SIP implementation documentation should state that it …

jjhenry21:
…MUST support only a limited set of file extensions

Added in the remarks section: "SIPs must support a limited set of file types and file extensions. The fileSupportedName function must check that the provided file matches one of the file extensions supported by the SIP.  For instance, the WSH SIP supports only the following list of file extensions and checks that the file under validation is a member of the following list: .js, .jse, .vbe, .vbs, or .wsf ".


PRMerger-2:
@jjhenry21 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #943, opened 2021-11-01
Update nf-virtdisk-setvirtualdiskmetadata.md

mattkur:
Fix typo (retrieve -> set).
Add a remark that there are max sizes. For example, the Windows VHDMP implementation currently enforces:
* 1MB per metadata entry
* 1024 metadata entries or 40MB of total metadata space (and holes in this space may lead to less actual capacity than advertised).

These sizes are implementation details and can change over time.

.vhd files don't support this sort of metadata.


PRMerger13:
@mattkur : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:

PR #949, opened 2021-11-05
Clarify out parameters of TryCreatePackageDependency and AddPackageDependency

sylveon:
The mention of `HeapAlloc` here is more confusing than anything, it implies that you have to heap allocate the memory passed here, when you don't: you pass a pointer to a variable and the function updates that variable, then you must deallocate the resource when done. Updated their description to make them more in line with similar out parameters in the Windows API, and clearer.


PRMerger12:
@sylveon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

asklar (approved):

PR #952, opened 2021-11-11
Updated the listed technologies used by a given header based on the respective config.json

tannergooding:
The index pages for many header files only listed one of the relevant technology areas they were used in, even when they were used by multiple.

This lead to various oddities like:
* `wingdi.h` being listed as `Data Exchange` rather than the more appropriate `Windows GDI`
  * https://docs.microsoft.com/en-us/windows/win32/api/wingdi/
  * https://docs.microsoft.com/en-us/windows/win32/api/_dataxchg/
  * https://docs.microsoft.com/en-us/windows/win32/api/_gdi/
* `winnt.h` being listed as `Backup` rather than the more appropriate `System Services`
  * https://docs.microsoft.com/en-us/windows/win32/api/winnt/
  * https://docs.microsoft.com/en-us/windows/win32/api/_backup/
  * https://docs.microsoft.com/en-us/windows/win32/api/_base/
* etc

I wrote a small tool to parse the `config.json` for each tech tree. This built a map of headers to technology areas and then I parsed the relevant `index.md` for the headers and updated to list the correct technology tree or all technology trees in which case multiple use the header:
* https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/_base/config.json
* https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/minwinbase/index.md

This results in many headers being updated to follow the format of one of the following.

### Where a single technology area uses the header
```
This header is used by Antimalware Scan Interface. For more information, see:

- [Antimalware Scan Interface](../_amsi/index.md)
```

### Where multiple technology areas use the header
```
This header is used by multiple technologies. For more information, see:

- [DirectShow](../_dshow/index.md)
- [Microsoft TV Technologies](../_mstv/index.md)
```

-------------------------

Notably there were about 348 header files listed in various tech areas that aren't available in this repo itself. Instead they are available in one of the other MicrosoftDocs repos. Some examples are headers that live in the Windows Driver Docs (80 headers) or Win32 Docs repos.
* https://github.com/MicrosoftDocs/windows-driver-docs-ddi
* https://github.com/MicrosoftDocs/win32

At least one of those headers was also a typo (`processthreadapi.h` should be `processthreadsapi.h`). I did not search the list to try and find other examples of these, etc.


sdk-api-src/content/accountssettingspaneinterop/index.md
```diff
@@ -15,6 +15,7 @@ f1_keywords:
 
 ## -description
 
-This header is used for WinRT Interoperability. For more information, see:
+This header is used by Windows Runtime C++ reference. For more information, see:
+
+- [Windows Runtime C++ reference](../_winrt/index.md)
```
tannergooding: There's a couple WinRT headers that used to link to a more specific file (`/uwp/api/Windows.UI.ApplicationSettings.AccountsSettingsPane` in this case).

That's, however, inconsistent with how all the other index docs link back to their tech tree. Perhaps the more direct link should be under a `See Also` section or something similar and with a different paragraph/link style so tooling can differentiate?

sdk-api-src/content/audioenginebaseapo/index.md
```diff
@@ -16,7 +16,8 @@ f1_keywords:
 
 ## -description
 
-This header is part of the Audio Devices DDI Reference. For more information, see:
+This header is used by multiple technologies. For more information, see:
```
tannergooding, line 19: Some of the paragraphs used to link to the relevant tech tree were inconsistent. For example, this one was `This header is part of the ...` rather than `This header is used by ...`. This change normalizes everything to be consistent on the latter as that was the most prevalent I found.

sdk-api-src/content/aclui/index.md
```diff
@@ -16,7 +16,8 @@ f1_keywords:
 
 ## -description
 
-This header is used by Security and Identity. For more information, see:
+This header is used by multiple technologies. For more information, see:
 
+- [CIMWin32 WMI Providers](../_cimwin32/index.md)
```
tannergooding, line 21: Notably on those that list multiple; the YML header still only lists a single `tech.root`. I'd imagine more complex changes are needed to properly link multiple?

tannergooding: I'd be happy to update that as well if needed, it'd just change from:

```
tech.root: security
```
to
```
tech.root:
 - cimwin32
 - security
```

sdk-api-src/content/windows.graphics.effects.interop/index.md
```diff
@@ -16,7 +16,8 @@ f1_keywords:
 
 ## -description
 
-This header is used by Windows Runtime C++ reference. For more information, see:
+This header is used by multiple technologies. For more information, see:
 
+- [w_graph_fx](../_w_graph_fx/index.md)
```
tannergooding, line 21: A couple technology areas (5 of them) don't have a "displayName":
* winlocation_com_ref
* winsensors_com_ref
* wlbsprov
* w_graph_fx
* w_ui_comp

drewbatgit:
@tannergooding - Hello! Thanks for your work on this, you did an incredible job fixing a problem that's been around for a long time. While almost everything you've done here is perfect, there are a couple of individual fixes you made that I think should be rolled back. I'll reply to each of your comments above individually. Thanks!

sdk-api-src/content/accountssettingspaneinterop/index.md
drewbatgit: @tannergooding For this one, the text calling out "WinRT Interoperability" was chosen deliberately to call out the specific developer scenario for these APIs. It's true that it's inconsistent with the other headers, but this was specifically requested by the PMs for the WinUI platform. Can you revert this change in just the *interop headers?

sdk-api-src/content/audioenginebaseapo/index.md
drewbatgit: @tannergooding  This is a good change, thanks.

sdk-api-src/content/aclui/index.md
drewbatgit: @tannergooding Unfortunately, our build system only supports associating a header with a single technology area in the YML header. Adding the links to all associated technologies in the content is a great improvement, but we need to keep the tech.root values pointing to a single technology.

tannergooding:
> While almost everything you've done here is perfect, there are a couple of individual fixes you made that I think should be rolled back. I'll reply to each of your comments above individually. Thanks!

Sure thing there were bound to be a couple mistakes with the automated script doing things 😄.

I'll get them addressed after you've finished identifying them.

sdk-api-src/content/accountssettingspaneinterop/index.md
tannergooding: Can do.

sdk-api-src/content/windows.graphics.effects.interop/index.md
drewbatgit: @tannergooding Thanks for calling these out. Some of these I have found names for, others I think I will remove. I'll go ahead and make these changes in our internal client repo so I can verify the fixes internally before merging the changes, but thanks again for identifying these.

tannergooding:
@drewbatgit, I believe I've addressed the actions items given. Please let me know if you need anything else on my end.

drewbatgit:
@tannergooding Looks great! Thanks for your contributions to the Windows docs!

PR #956, opened 2021-11-11
Improving a link in D3D12 documentation

hannes-harnisch:
The struct that this doc page is describing is used only for a specific function on a specific COM interface, but the existing link doesn't lead to that function's documentation (despite the text implying that) and also spells the interface's name wrong.


PRMerger10:
@hannes-harnisch : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #972, opened 2021-11-22
fix parameter name in RegisterHotKey documentation

mikeblas:
Another parameter mentions "uVirtKey", but the parameter was defined
as "vk" here.


PRMerger20:
@mikeblas : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mikeblas:
How do I move this forward?

jwmsft:
Hi @mikeblas, thanks for taking the time to submit this PR. Unfortunately, we can't accept this update. The parameter name is defined as `vk` in WinUser.h, and the documentation has to match that in order to build correctly. If we change the parameter name in the documentation, it will cause errors in the doc build.

PR #979, opened 2021-11-26
Corrections and clarifications for FILE_RENAME_INFO

idigdoug:
Remove duplicated ReplaceIfExists field.

Since ReplaceIfExists and Flags are in a union, we need to explain when each one is used. ReplaceIfExists is used for FileRenameInfo, and Flags is used for FileRenameInfoEx.

Ideally the accepted values for Flags should be documented (they're listed in https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_rename_information), but I'm not sure which flags should be shown here or the version of windows when each flag becomes valid, so I'm not going to add that.

Clarify the usage of RootDirectory. Most users should set it to NULL. Wording is taken from https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_rename_information.

FileNameLength needs to be clearly documented as the FileName size in bytes without the NUL termination. (Technical note: at present this value is only used in some code paths. In other code paths, FileName needs to be nul-terminated.)

Clarify that FileName MUST be NUL-terminated. Some code paths use the NUL-termination and other code paths use FileNameLength, so both must be set correctly.

### DIFF
--- a/sdk-api-src/content/winbase/ns-winbase-file_rename_info.md
+++ b/sdk-api-src/content/winbase/ns-winbase-file_rename_info.md
@@ -73,8 +73,7 @@ This field is used when **SetFileInformationByHandle**'s *FileInformationClass* 
 
 ### -field RootDirectory
 
-If the file is not being moved to a different directory, or if the *FileName* member contains the full path to the target file, this member is
-**NULL**. Otherwise, it is a handle for the root directory under which the file will reside after it is renamed.
+This field should be set to NULL.
 
 ### -field FileNameLength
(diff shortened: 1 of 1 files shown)


PRMerger4:
@idigdoug : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:
@idigdoug : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #984, opened 2021-12-01
Update nf-netfw-networkisolationenumappcontainers.md

ChrisOfTheLewi:
We had a software vendor ask about this, and it isn't clear to me from the current doc description what to do.  Looking at our current implementation, it appears that NetworkIsolationFreeAppContainers would be the correct call for freeing the memory.  Not freeing it in any way results in a memory leak.


PRMerger13:
@ChrisOfTheLewi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #985, opened 2021-12-01
Fixing a few problematic yaml headers

tannergooding:
A few of the YAML headers were "invalid" leading to problems when attempting to parse them using tools like YamlDotNet.

### DIFF
--- a/sdk-api-src/content/aviriff/ns-aviriff-avistreamheader.md
+++ b/sdk-api-src/content/aviriff/ns-aviriff-avistreamheader.md
@@ -2,7 +2,7 @@
 UID: NS:aviriff._avistreamheader
 title: AVISTREAMHEADER (aviriff.h)
 description: The AVISTREAMHEADER structure contains information about one stream in an AVI file.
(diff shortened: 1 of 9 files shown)


PRMerger15:
@tannergooding : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
I don't know what the valid format(s) of helpviewer_keywords is/are. Perhaps Drew can confirm that the changes here are good.

tannergooding:
> I don't know what the valid format(s) of helpviewer_keywords is/are

I had looked at a few example files and there isn't a whole lot of consistency.

`helpviewer_keywords` is generally an abbreviated list: `: [ item1, item2, item3 ]`
while `ms.keywords` is generally a value of the same terms `: item1, item2, item3`

They are sometimes quoted and sometimes not (but seeing as its YAML that generally shouldn't matter too much, as its a string regardless and the quotes are just used for escaping and such)

-----------------------

On a meta note, there are also tons of markdown errors that prevent the files from being parsed by any CommonMark compliant parser; many of these errors are due to things like unbalanced HTML tags, unnecessary newlines causing an HTML block to be terminated too early, or missing newlines causing what should have been a separate paragraph block to instead be treated as part of an HTML block.

These issues surface in a myriad of ways on the docs pages, such as:
image

I'd like to submit a PR for these, but such a PR would be fairly massive and would touch at least 25k files last I checked. It would be nice if some solution could be found for this and I'm happy to submit said PR if appropriate.

I did write a tool which largely processes and handles these issues, so its fairly automated overall. Happy to have someone reach out internally on teams if that's appropriate and maybe we can work something out. Alias is `tagoo`

MatchaMatch:

PR #987, opened 2021-12-02
Correct minimum supported Build

vclin0:
IMFExtendedCameraController - Win32 apps | Microsoft Docs is available in 19041+, while the MSDN doc says it is available only Build 20348 and up.


PRMerger7:
@vclin0 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #993, opened 2021-12-09
Update nf-memoryapi-mapviewoffile2.md

walbourn:
Couldn't find any DLL which contains MapViewOfFile2 to get from GetProcAddress. Turns out it's not an exported function.

Also updated the links to current location.


PRMerger13:
@walbourn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

glenn-slayden:
I believe `MapViewOfFile2` is a compile-time alias for calling `MapViewOfFileNuma2` with `NUMA_NO_PREFERRED_NODE = 0xFFFFFFFF` as the last parameter?

walbourn:
Yup, that's what my edit is for. If you try to find it via ``GetProcAddress``, it doesn't exist.

Karl-Bridge-Microsoft (approved):

PR #995, opened 2021-12-13
Fix minimum OS version for IsWow64Process2

rickbrew:
The docs have been wrong for years. Win10 v1511 is not the minimum OS, it is Win10 v1709.

See also dotnet/runtime#26612 (comment)

> I think the documentation doesn't correctly reflect the version where this function is available; IsWow64Process2 is actually only available starting with RS3 (Windows 10 Version 1709, Build 16299).


PRMerger3:
@rickbrew : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #997, opened 2021-12-15
[SetFileTime] Remove unclear duplicated information

glenn-slayden:
The exact same text from `lpLastAccessTime` is repeated verbatim for `lpLastWriteTime`.

Since the text in question specifically only mentions `lpLastAccesTime`, it is not clear whether the special input condition applies to the `lpLastWriteTime` parameter or not.

### DIFF
--- a/sdk-api-src/content/fileapi/nf-fileapi-setfiletime.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-setfiletime.md
@@ -94,12 +94,6 @@ To prevent file operations using the given handle from modifying the last access
 A pointer to a <a href="/windows/desktop/api/minwinbase/ns-minwinbase-filetime">FILETIME</a> structure that 
        contains the new last modified date and time for the file or directory.  If the application does not need to change this information, set this parameter either  to <b>NULL</b>  or to a pointer to a <b>FILETIME</b> structure that has both the <b>dwLowDateTime</b> and <b>dwHighDateTime</b> members set to 0.
 
-To prevent file operations using the given handle from modifying the last access time, call 
-       <b>SetFileTime</b> immediately after opening the file handle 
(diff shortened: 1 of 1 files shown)


PRMerger15:

sdk-api-src/content/fileapi/nf-fileapi-setfiletime.md
```diff
@@ -94,12 +94,6 @@ To prevent file operations using the given handle from modifying the last access
 A pointer to a <a href="/windows/desktop/api/minwinbase/ns-minwinbase-filetime">FILETIME</a> structure that 
        contains the new last modified date and time for the file or directory.  If the application does not need to change this information, set this parameter either  to <b>NULL</b>  or to a pointer to a <b>FILETIME</b> structure that has both the <b>dwLowDateTime</b> and <b>dwHighDateTime</b> members set to 0.
 
-To prevent file operations using the given handle from modifying the last access time, call
```
alvinashcraft, line 97: I checked with the product team, and this paragraph should be kept. We should just update "last access time" to "last write time" in this existing paragraph.

alvinashcraft (changes requested):
Paragraph should be updated instead of removed, according to the product team. See my comment inline. Thanks for finding this issue!

alvinashcraft:
Hi, @glenn-slayden. I'm checking in to see if you had a chance to look at the proposed change to this PR based on the product team's suggestion? Do you want to move forward with the updated paragraph, or should we close this PR and handle it separately? Thanks again for your contribution and for bringing this issue to our attention.

glenn-slayden:
@alvinashcraft  Please go ahead and make the changes in any way you see fit. I do not need to supervise the PR. Thanks,

Glenn

alvinashcraft:
Hi @glenn-slayden. I just checked in the changes to our internal repo. It should sync to the public repo and the live site shortly. Thanks again for your contribution here. I'm going to close this PR, but feel free to circle back if you have any comments.

PR #1014, opened 2022-01-03
Mention that POINTER_INFO dwKeyStates is always 0

lb90:
Hi! It seems that dwKeyStates is always 0 regardless of the key state, as such I think it should be documented. Let me know 🙂


PRMerger8:
@lb90 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
Removed the "unused" statement and added a link to the relevant constant topic.

The suggested revision contradicted the original description, which already states the value may be zero. 

Please open a new PR or Issue with more detail about the pointer devices you were using and how. 

Thanks!

PR #1015, opened 2022-01-04
Improve parameter names for IDirectDraw7::CreateClipper

clandrew:
For this page:
https://docs.microsoft.com/en-us/windows/win32/api/ddraw/nf-ddraw-idirectdraw7-createclipper

Something systemic took these parameters out, see here where we used to have them
https://docs.microsoft.com/en-us/previous-versions/ms785026(v=vs.85)

I don't know how to fix the systemic issue, this fixes the issue on this one page.


PRMerger-2:
@clandrew : Thanks for your contribution! The author(s) have been notified to review your proposed change.

clandrew:
I'm abandoning this since I would prefer a fuller fix instead

PR #1017, opened 2022-01-04
Fix escaped backslashes

Skinner927:
There are undoubtedly more instances of this erroneous escape happening in other documents.
Oddly, the same strings inside the `<div class="alert">` boxes do not have this problem.


PRMerger10:
@Skinner927 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1046, opened 2022-01-31
Correct parameter names for WndProc callback func

ffio1:
All the parameter names were "unnamedParam#". This can make the names confusing for new users, especially when the description for each parameter says what the name is typically.


PRMerger6:
@ffio1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft:
Hi, thanks for taking the time to submit this PR. Unfortunately, we can't accept this update. Our doc build system uses the `unnamedParam#` convention for any parameters that aren't named in the header file. If we change the parameter name in the documentation, it will cause errors in the doc build. In this case, the typical name is included in the parameter description instead.

PR #1047, opened 2022-01-31
Remove obsolete _MAC fields

malxau:
The three fields at the end of `WIN32_FIND_DATA` are under an `#ifdef _MAC`.  This was used 25 years ago to support Visual C++ Cross Development Edition For Macintosh, which was not carried forward into OS X so has been obsolete for at least 20 years.  Anyone targeting Windows will not have these fields present in the structure, and suggesting they do is misleading.

This also removes the second definition of reparse tags from `WIN32_FIND_DATAW`.  Note the reparse tags are still included in the text above, and this second definition is not present in `WIN32_FIND_DATAA`, so it looks to be mistakenly included.

### DIFF
--- a/sdk-api-src/content/minwinbase/ns-minwinbase-win32_find_dataa.md
+++ b/sdk-api-src/content/minwinbase/ns-minwinbase-win32_find_dataa.md
@@ -187,12 +187,6 @@ An alternative name for the file.
 
 This name is in the classic 8.3 file name format.
 
-### -field dwFileType
-
-### -field dwCreatorType
-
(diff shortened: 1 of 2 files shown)


PRMerger12:
@malxau : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
Hi @malxau - These fields are still in the header and, as such, still need to be included in the documentation. There's also the corresponding W struct that needs the same update. I'll close this PR and spin up a private one with the suggested info. Thanks.

PR #1073, opened 2022-02-22
TextOutW: Fix argument name, from "c" to "cchString"

diamondo25:
The argument "lpString" refers to "cchString" as the argument of the string length, but did not exist. However, there is an argument matching the explanation, which was just called "c". Therefor, changing "c" to "cchString" should fix the API definition.


PRMerger-2:
@diamondo25 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

diamondo25:
TextOutA has the same error, and should most likely be changed too. Please advice.

GrantMeStrength:
Currently we're unable to use better param names in the docs, as it breaks rendering.

PR #1083, opened 2022-03-07
Clarify wording on lpValue

henke37:
The word "must" makes it clear that it is a requirement and not a mere recommendation. Also make it clear that it's the list that is being referred to.


PRMerger3:
@henke37 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1092, opened 2022-03-15
Clarified that explanation of "which value is used as nCmdShow" requires WS_VISIBLE and x == CW_USEDEFAULT

Wallby:
I think that this is what is meant. I was not entirely sure reading the current paragraph. I think it is not deducable from reading this specific paragraph alone that "_If the y parameter is CW_USEDEFAULT, then the window manager calls ShowWindow with the SW_SHOW flag after the window has been created_" and "_If the y parameter is some other value, then the window manager calls ShowWindow with that value as the nCmdShow parameter_" apply to if WS_VISIBLE and x == CW_USEDEFAULT.


PRMerger7:
@Wallby : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger18:
@Wallby : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1094, opened 2022-03-15
Update ns-winuser-rawinputheader.md - Precision touchpad remark (confusing hDevice == 0)

niello:
The problem and the answer are described here:
https://stackoverflow.com/questions/57552844/rawinputheader-hdevice-null-on-wm-input-for-laptop-trackpad

Just encountered the same problem with my new notebook. If someone has more knowledge on topic please write more elaborate remark.


PRMerger20:
@niello : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1096, opened 2022-03-16
Update nf-shlwapi-pathisfilespecw.md

compnerd:
Escape a couple of instances of `\` as the character itself is an escape character and is not rendered without escaping.


PRMerger10:
@compnerd : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1097, opened 2022-03-17
Update casing for implied extension.

AaronRobinsonMSFT:
This clarifies the casing for the implied ".DLL" extension that the loader will apply.

Recommend viewing with "ignore whitespace" - https://github.com/MicrosoftDocs/sdk-api/pull/1097/files?diff=split&w=1

### DIFF
--- a/sdk-api-src/content/libloaderapi/nf-libloaderapi-loadlibrarya.md
+++ b/sdk-api-src/content/libloaderapi/nf-libloaderapi-loadlibrarya.md
@@ -90,7 +90,7 @@ If the function cannot find the  module, the function fails. When specifying a p
 
 If the string specifies a module name without a path and the file name extension is omitted, the function
        appends the default library extension ".DLL" to the module name. To prevent the function from appending
-       .dll to the module name, include a trailing point character (.) in the module name string.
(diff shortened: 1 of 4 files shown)


PRMerger10:
@AaronRobinsonMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

GovM-MSFT (commented):
Do we want to leave a note about case sensitivity at the filesystem level at all? Like it is inadvisable to enable case sensitivity on any location in %programfiles% or to any directory that you are loadings modules from.

AaronRobinsonMSFT:
> Do we want to leave a note about case sensitivity at the filesystem level at all? Like it is inadvisable to enable case sensitivity on any location in %programfiles% or to any directory that you are loadings modules from.

Based on an offline conversation, we've decided the existing generic case-sensitive warning at [here](https://docs.microsoft.com/windows/wsl/case-sensitivity#change-the-case-sensitivity-of-files-and-directories), might be sufficient. I defer to the documentation owners if they believe otherwise.

PRMerger18:
@AaronRobinsonMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1098, opened 2022-03-18
Updating ACLineStatus field

maserwin:
Update ns-winbase-system_power_status.md so the possible ACLineStatus field members match internal winbase.h


PRMerger12:
@maserwin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1102, opened 2022-03-22
Line 63 updated

jjhenry21:
Changed “The following error status codes are defined for certificates and chains.”  to “dwErrorStatus is a bitmask of the following error codes defined for certificates and chains.”


PRMerger13:
@jjhenry21 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1124, opened 2022-04-08
Update nf-gdiplusmatrix-matrix-matrix(real_real_real_real_real_real).md

fdwr:
Current descriptions are tautologically non-descript (e.g. m11 is the first row, first column) rather than explaining what they actually mean. See the descriptions in GDI's [XFORM](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-xform) and [DWRITE_MATRIX](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/ns-dwrite-dwrite_matrix) for comparison, which are both bitwise compatible with the GDI+ equivalent.

(*Just adding this because I had to look up the definitions myself and couldn't recall which was which 😅, whether it was m12 or m21 which was the negative sine.)


PRMerger19:

stevewhims (approved):

PRMerger20:

PR #1127, opened 2022-04-12
remove "[desktop apps only]" from RegGetValue docs

ChrisGuzak:
The APIs are available on all windows editions.

“desktop apps only” means the APIs use is blocked by not making the API definitions available when compiling a UWP App project (see WINAPI_FAIMILY_PARTITION stuff below) and by verifying that binaries, submitted to the Windows Store did not call these APIs (this is known as the WACK).

On this API we changed this policy and allow use from apps based on the use of `WINAPI_PARTITION_APP`

winbase.h from Windows 11 SDK (22000)
```cpp
#pragma region Application Family or OneCore Family
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)

…

WINADVAPI
LSTATUS
APIENTRY
RegGetValueW(
    _In_ HKEY hkey,
```

### DIFF
--- a/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
+++ b/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
@@ -11,8 +11,8 @@ ms.keywords: RRF_NOEXPAND, RRF_RT_ANY, RRF_RT_DWORD, RRF_RT_QWORD, RRF_RT_REG_BI
 req.header: winreg.h
 req.include-header: Windows.h
 req.target-type: Windows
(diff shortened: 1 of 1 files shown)


PRMerger9:
@ChrisGuzak : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ChrisGuzak:
@stevewhims and other doc owners. this is only one of many APIs that need to be updated since we've changed the policy on which APIs UWP apps can use.

PR #1131, opened 2022-04-18
update return value of Free

janasebaali:
update return value of NetworkIsolationFreeAppContainers to match code


PRMerger3:
@janasebaali : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1134, opened 2022-04-21
Clarify ELEMENT_WISE_CLIP casting behavior

fdwr:
For https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_element_wise_clip_operator_desc

If the tensor data type is not *float*, then *Min* and *Max* are cast to the tensor data type before applying the clipping operation, which for integers means truncating toward zero and for floating point values rounding to nearest even.


PRMerger12:

stevewhims (approved):

PRMerger5:

PR #1145, opened 2022-04-28
Remove silly WPSTYLE_MAX remark

sredna:
You can't set it to WPSTYLE_MAX because it 
1) is not the same as WPSTYLE_SPAN.
2) is not even a valid value, it is only for range checks.


PRMerger12:
@sredna : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1148, opened 2022-04-30
Fix the sample code

gvanem:
Make the example compile. The error from 'cl' was:
```
error C2440: 'function': cannot convert from 'GUID' to 'const GUID *const '
warning C4024: 'StringFromGUID2': different types for formal and actual parameter 1
```


PRMerger14:
@gvanem : Thanks for your contribution! The author(s) have been notified to review your proposed change.

gvanem:
I now see the original code is correct for C++. But not for plain C.

PR #1150, opened 2022-05-03
Document behavior of GetWindowRect() on minimized windows

fredemmott:
It's surprising to me that calling GetWindowRect() on a minimized window is not a failure - instead, it sets a bogus rect. Document this in remarks.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-getwindowrect.md
+++ b/sdk-api-src/content/winuser/nf-winuser-getwindowrect.md
@@ -97,6 +97,8 @@ Calling GetWindowRect will have different behavior depending on whether the wind
 
 To get the window bounds excluding the drop shadow, use <a href="/windows/win32/api/dwmapi/nf-dwmapi-dwmgetwindowattribute">DwmGetWindowAttribute</a>, specifying <b>DWMWA_EXTENDED_FRAME_BOUNDS</b>.  Note that unlike the Window Rect, the DWM Extended Frame Bounds are not adjusted for DPI.  Getting the extended frame bounds can only be done after the window has been shown at least once.
 
+If the window is minimized (iconic), GetWindowRect will return a nonzero value, and the top left corner of the rect will be (-32000, -32000).
+
 
 #### Examples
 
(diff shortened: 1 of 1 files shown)


PRMerger-2:
@fredemmott : Thanks for your contribution! The author(s) have been notified to review your proposed change.

DJm00n:
@fredemmott I think documenting `-32000` value is a bad idea. Better to describe this behaviour something like this:
 
`GetWindowRect() behaviour is undefined for minimized windows. You can use to IsIconic() function to check that window is minimized.`

This comply with @oldnewthing's  [Where did windows minimize to before the taskbar was invented?](https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453) and [Why do minimized windows have an apparent size of 160×31?](https://devblogs.microsoft.com/oldnewthing/20050210-00/?p=36483) blog posts and [comment on StackOverflow](https://stackoverflow.com/questions/37676311/get-windows-current-dimensions-regardless-of-minimise-normal-maximise-state#comment62840181_37677375).

fredemmott:
Even if the -32000 value is not explicitly documented, it would be good to explicitly document that even if undefined, it 'may' return `true`, despite returning an invalid value.

Documenting the -32000 value is useful though: I came across this behavior when several users reported that the window was 'invisible', and the app had saved (-32000,-32000) as the window position to restore on next launch. Documenting this value would have helped debug it - perhaps it should be documented, but not as guaranteed behavior?

DJm00n:
>it 'may' return true, despite returning an invalid value.

Yes, exact wording is up to you. I just suggested my variant.

>Documenting the -32000 value is useful though

`(-32000,-32000)` is CURRENT behavior for minimized windows that already have changed several times in Windows life. Documenting current behaviour is a bad thing in long run - it could cause even more compatibility issues in the future.

We need to discourage use of `GetWindowRect()` if window is minimized (use `IsIconic()` call to check) and saving of negative rect coords in configs. :)

oldnewthing:
We should document that the window rect for a minimized window may be off-screen, but should not document the precise values because they do vary over time. Furthermore, the window rect for a minimized window is usually **but not always** off-screen, so we can't even promise that they are off-screen, much less at a specific coordinate.

> It is common for the Windows shell to move minimized (iconic) windows off-screen. Programs should be careful not to apply minimized window coordinates to restored or maximized windows.

PRMerger4:
@fredemmott : Thanks for your contribution! The author(s) have been notified to review your proposed change.

fredemmott:
> discourage ... saving of negative rect coords in configs

I agree on this point - and perhaps documenting the minimized behavior in remarks is sufficient. On the other hand, without this note, my initial perspective was trying to figure out *why* my users had -32000 in their saved configurations - minimization was a pure guess. My app does now check `IsIconic()` and prevents it happening in the future, but that doesn't change the initial debugging perspective if others end up in a similar situation.

fredemmott:
Closing, no response from MS in ~ 2 years.

PR #1157, opened 2022-05-11
Deprecate OpenIcon and CloseWindow functions

DJm00n:
These two methods are still here only for compatibility reasons.

See @oldnewthing's [Where did windows minimize to before the taskbar was invented?](https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453) and [Why do minimized windows have an apparent size of 160×31?](https://devblogs.microsoft.com/oldnewthing/20050210-00/?p=36483) blog posts.

Related #1150.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-closewindow.md
+++ b/sdk-api-src/content/winuser/nf-winuser-closewindow.md
@@ -48,6 +48,8 @@ req.apiset: ext-ms-win-ntuser-window-l1-1-4 (introduced in Windows 10, version 1
 
 # CloseWindow function
 
+> [!IMPORTANT]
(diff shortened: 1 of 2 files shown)


PRMerger8:
@DJm00n : Thanks for your contribution! The author(s) have been notified to review your proposed change.

oldnewthing:
I disagree. The functions are still working and supported, they are not on any deprecation path. They are an old-fashioned way of doing things, but old-fashioned doesn't mean "not recommended". They are also subtly different from ShowWindow(): If you try to OpenIcon a maximized window, nothing happens. But ShowWindow(SW_SHOWNORMAL) will change it to restored.

DJm00n:
@oldnewthing Thank you for the comment!

So `OpenIcon()` is something like this:

```cpp
if (IsIconic(hWnd))
    ShowWindow(hWnd, SW_SHOWNORMAL);
```

Closing then.

PR #1158, opened 2022-05-11
Clarify recommendation in CloseThreadpoolTimer and CloseThreadpoolWait.

bentonj-vmw:
It seems that a 0 duration set must be followed by a wait with cancel as TRUE in order to actually ensure that no further callbacks are called.

We ran into this bug multiple times where by using FALSE we were still getting the occasional unexpected callback, so I think clarifying the documentation here will be of great help to any future readers..


PRMerger13:
@bentonj-vmw : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1162, opened 2022-05-15
Update nf-fileapi-writefile.md

defrag257:
In Windows 7 and earlier, if lpNumberOfBytesWritten and lpOverlapped are both NULL, the program will crash.
In Windows 8 and later, if lpNumberOfBytesWritten and lpOverlapped are both NULL, the program will run as normal.
Why Windows 8 added this support but not documented is a mystery.
Probably because lpNumberOfBytesWritten in synchronous writes is duplicated as all bytes must be written in a successful synchronous write.
Formerly merged PR "Windows 7:  This parameter can not be NULL." is not correct.
In fact, in Windows 7, when lpOverlapped is not NULL, this parameter (lpNumberOfBytesWritten) CAN be NULL as usual.


PRMerger17:
@defrag257 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1168, opened 2022-05-20
Update the document to reflect actual annotations

hwisungi:
In the library header winver.h, this function has _Reserved_ annotation on the third "dwHandle" parameter, requiring the value to be NULL or 0(zero). But currently this document indicates that the value will be ignored instead. Either update the document to reflect the current annotation on the parameter, or have the owner of the header to update the annotation to reflect the actual behavior if the function really ignores this value.


PRMerger-2:
@hwisungi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1169, opened 2022-05-20
pathcch.h functions out of date and missing information

rseanhall:
This change isn't intended to be merged as is, but I could not find the correct place to create an issue for this documentation.

All of the pathcch.h functions are missing the required information to delay load them from the desktop. In addition, it appears that the documentation for the flags are out of date with the header, as you can see by the missing `PATHCCH_CANONICALIZE_SLASHES` value.

After further reviewing the file, it appears that the delay load information might be there but is not rendered properly -  I needed to know the information in the header in `api_location` - `API-MS-Win-Core-Path-l1-1-0.dll`/`KernelBase.dll`.


PRMerger13:
@rseanhall : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1187, opened 2022-06-08
Windows 7 requires lpNumberOfBytesRead on ReadFile to be set

GXTX:
There is an undocumented quirk existing on Windows 10 & 11 where you can use `WriteFile` and `ReadFile` without setting  `lpNumberOfBytesWritten` & `lpOverlapped`.

We ran into a bug running code on Windows 7 where it actually *must* be set.
https://github.com/GXTX/YACardEmu/commit/7e9d80b3f2

The page for `WriteFile` already warns about this behavior, so the fix here is to warn on `ReadFile` as well.


PRMerger20:
@GXTX : Thanks for your contribution! The author(s) have been notified to review your proposed change.

alvinashcraft (approved):
Nit: "cannot" should be one word, but it does match the wording in the WriteFile docs as it is now. Thanks for catching this and submitting the fix.

alvinashcraft:
#sign-off

PR #1206, opened 2022-06-20
Document that `FILE_NAME_INFO` may be modified on failure

ChrisDenton:
The `FileNameLength` field of [`FILE_NAME_INFO`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_name_info) may be modified if `GetFileInformationByHandleEx` fails. Documenting this is important for two reasons:

1. it enables handling longer paths without needing to guess the length
2. it may be important to note that the `FileNameLength` field can, in some circumstances, contain a byte length larger than the currently available buffer

### DIFF
--- a/sdk-api-src/content/winbase/ns-winbase-file_name_info.md
+++ b/sdk-api-src/content/winbase/ns-winbase-file_name_info.md
@@ -68,6 +68,12 @@ The size of the <b>FileName</b> string, in bytes.
 
 The file name that is returned.
 
+## -remarks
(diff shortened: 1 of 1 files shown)


PRMerger5:
@ChrisDenton : Thanks for your contribution! The author(s) have been notified to review your proposed change.

alvinashcraft (approved):
Thank you for updating the documentation to explain this scenario.

PR #1226, opened 2022-07-06
Update nf-synchapi-sleepex.md

jeffbromberger:
In the description of the dwMilliseconds parameter, the mention of pending APCs combined with the bAlertable parameter set to FALSE was misleading.  When bAlertable is FALSE, APCs will not be delivered.  I.e. the result is the same whether an APC is queued or not - no APC callback will execute.  To say "if there are no pending user APCs" implied that this condition would somehow impact the outcome.

In the bAlertable description, I was initially confused by the phrase "the function does not return".  I now understand that it means "does not return immediately".  I added those words for clarity, but leave it at your discretion as to whether that is needed.

Small typo correcting "timer-out" to "time-out"

### DIFF
--- a/sdk-api-src/content/synchapi/nf-synchapi-sleepex.md
+++ b/sdk-api-src/content/synchapi/nf-synchapi-sleepex.md
@@ -69,7 +69,7 @@ Suspends the current thread until the specified condition is met. Execution resu
 
 The time interval for which execution is to be suspended, in milliseconds.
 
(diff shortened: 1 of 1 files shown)


PRMerger10:
@jeffbromberger : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
Thanks @jeffbromberger!

PR #1236, opened 2022-07-13
Update nf-unknwn-iunknown-queryinterface(refiid_void).md

colejohnson66:
Correct `void` to `void**` in title

### DIFF
--- a/sdk-api-src/content/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void).md
+++ b/sdk-api-src/content/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void).md
@@ -1,13 +1,13 @@
 ---
-UID: NF:unknwn.IUnknown.QueryInterface(REFIID,void)
-title: IUnknown::QueryInterface(REFIID,void)
+UID: NF:unknwn.IUnknown.QueryInterface(REFIID,void**)
+title: IUnknown::QueryInterface(REFIID,void**)
 description: Retrieves pointers to the supported interfaces on an object.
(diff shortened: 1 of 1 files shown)


PRMerger6:
@colejohnson66 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, @colejohnson66! Changing this is a nice thought in principle. But here are the reasons why we don't need to (or mustn't) make the changes in this PR.

**UID**. We must never make any change to the value of the UID field. If we do, then the topic disappears from the live site. And the only fix for that is to change the UID back to what it was. :) It's best to think of the UID field as an opaque identifier; it's never seen by readers of the topic (only by its authors).

**title**. The value of this field is displayed only in a browser tab (and in a tooltip on mouseover). And with the browser I use, the title is already truncated so that you can't even see the end of the word "void". So appending "**" to that won't be seen either. In general, I think we should leave the title field alone. It's not cost-effective to invest in working on it: I feel that the auto-gen'd value will be sufficient for every imaginable use case. Btw, the actual topic title that you see in large bold font in the browser is auto-gen'd, and doesn't itself appear in the content (even if a markdown H1 title is present in the topic, that's not the value that's rendered on the page). :)

**helpviewer_keywords**. Changing this would be harmless, but I doubt that change will be visible to, nor affect, anyone.

**ms.keywords**. In fact, this field shouldn't be in this topic. It's redundant, and harmless, so it's ok to leave it (but not invest in changing it).

colejohnson66:
Makes sense. However, re: the title, [on the live page][0], it is shown, and it's missing the double pointer stars. That's what prompted the change in the first place.

image

[0]: https://docs.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void)

PR #1238, opened 2022-07-18
HRESULT_FROM_WIN32: Replace HTML escapes in the example

Bryce-MW:
I noticed that this example has some HTML escapes left over so I removed them so that the example renders properly.


PRMerger16:
@Bryce-MW : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1242, opened 2022-07-22
fix typo in file name

wheybags:
This typo causes a broken link at https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-cd3d11_sampler_desc

Screenshot:
image


PRMerger15:
@wheybags : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, @wheybags. We don't want to rename files, if we can help it (and there can't be a typo in a filename, since those names are auto-generated by our content build system). I'd be more inclined to change the url in whatever link is broken. But I tried all the links in that topic and I didn't see any 404. Could you identify exactly which link is broken, and perhaps propose a pull-request that fixes the url itself?

Thanks!
-Steve

wheybags:
Hi @stevewhims
So, if you look at the screenshot I sent, you'll see a big garbled bunch of text wrapped in [] and (). That's because the markdown generated as a link to the ` explicit CD3D11_SAMPLER_DESC(D3D11_FILTER filter, D3D11_TEXTURE_ADDRESS_MODE ...` large constructor has unbalanced parentheses in it, which breaks the markdown html generation. At the end of the filename is an `_`, where there should be a `)` to close the parameter list. As a result you get that nasty text instead of a proper `<a>` tag. I presumed it was a typo because `_` and `)` are so close to each other on a keyboard.

I can reproduce the issue right here in comment markdown:

#### Correct
```Source md: [asd](a/(b))```
[asd](a/(b))

#### Broken
```Source md: [asd](a/(b_)```
[asd](a/(b_)

The broken link is [here](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-cd3d11_sampler_desc), but when I look at [the source for that article](https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/d3d11/ns-d3d11-cd3d11_sampler_desc.md), the whole section listing constructors is missing, so I presumed it was auto-generated somehow. 

Regardless, the file name is clearly a mistake (the pages for the other constructors have balanced parentheses, eg [this one](https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/d3d11/nf-d3d11-cd3d11_sampler_desc-cd3d11_sampler_desc(constd3d11_sampler_desc_).md)), and can't be linked to from markdown normally (you can escape the `(`, and then it will work I guess).

stevewhims:
Thanks, @wheybags, I see the rendering issue now. The content build system generates file names, and as you say it's also responsible for generating that chunk of content. I'm looking into whether it's possible to escape that first parenthesis and thereby make the link work. So the solution here will be for the build system to either a) properly escape the names of files, or b) generate file names that don't need escaping. I've passed that info along to the engineers in the form of a bug report. But it's not something that we can fix manually by changing content (or by manually renaming files).

Thanks again!
-Steve

wheybags:
Makes sense, thanks for your time!

PR #1245, opened 2022-07-26
Add ISwapChainPanelNative::SetSwapChain remarks

Sergio0694:
This PR adds the following remarks to `ISwapChainPanelNative::SetSwapChain`:
- Calls out the API has to be called from the right UI thread.
- Mentions the fact that `SetSwapChain` will add extra references to the target device, and that calling `SetSwapChain(null)` is needed in order to immediately release them when the panel is no longer needed (eg. to recover from a device lost scenario).

It also adds `[opt]` to the `swapChain` param.

This has tripped me up in ComputeSharp ([see this commit](https://github.com/Sergio0694/ComputeSharp/pull/319/commits/dbb28c35ce68eba7b1879a24fe11166a9b9b90b6)), as the target device was left with way more references than expected, as I wasn't calling `SetSwapChain(null)`. This seems like a rather important aspect of the API for those wanting to recover from device lost scenarios, and it should be properly documented. Also added the remarks on the UI thread while at it 🙂


PRMerger14:
@Sergio0694 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

GrantMeStrength:
Thanks for taking the time to improve the docs!

PR #1246, opened 2022-07-26
Update nf-unknwn-iunknown-queryinterface(refiid_void).md

colejohnson66:
Add double pointer stars to `void**` parameter.

### DIFF
--- a/sdk-api-src/content/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void).md
+++ b/sdk-api-src/content/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void).md
@@ -1,6 +1,6 @@
 ---
 UID: NF:unknwn.IUnknown.QueryInterface(REFIID,void)
-title: IUnknown::QueryInterface(REFIID,void)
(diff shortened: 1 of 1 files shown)


PRMerger7:
@colejohnson66 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, @colejohnson66, and thanks for the PR.

The value that you changed in this PR determines *only* what appears in the tab of the browser. Changing that field won't affect the H1 title as rendered (in large font) at the head of the web page. And it doesn't seem enough to change the former without changing the latter. And I don't think we should change the former by manually editing YAML like this, because that won't scale; instead we should ask engineering to generate a different value.

The H1 title on the web page is auto-generated by the content build system. It doesn't appear in the content of the topic, so writers can't change it.

So what I've done here is I've filed a work item on the content engineering team about fixing the type in both titles (the one in the web page, and the one in the title: field). I'll track that work item internally, so I'll close this PR, but I've added a link to it in the work item I created.

Thanks!
-Steve

stevewhims:
UPDATE: looks like the work item that I mention above is fixed.

PR #1261, opened 2022-08-03
Trying to fix wording of what header to include [WIP]

exoosh:
- The wording for which header to include is very awkward for a number of functions (I counted 162 occurrences overall). It seems to suggest that on older Windows versions you ought to include Windows.h but on newer ones you can use the header in which it is regularly declared.
- Since the string from `req.include-header:` appears to go into parentheses and gets an "include " slapped in front, this should fix the wording up a bit.

Regular expression used:
* search: `(req\.include-header:)\s+?(Windows .+?),\s+?([a-zA-Z0-9]+?\.h)`
* replace: `\1 \3 on \2`

This is the first part and still WIP. Since you don't allow issues, but only pull requests, it's the best I could come up with to get a discussion started on this awkward wording and how to fix it.

Best regards,

Oliver


exoosh:
The second commit was using the following regex:

* search: `(req\.include-header:)\s+?(Windows(?!\.h|\.UI\.|x\.h|\.Foundation\.).+), (\w+?\.h)`
* replace: `\1 \3 on \2`

drewbatgit:
Thanks for your work on this! This is an issue that has been on our back burner for a long time, so it's great to have it finally resolved.

PR #1262, opened 2022-08-05
Fix dml_batch_normalization_grad_operator note on broadcasting size 1 dimensions

fdwr:
In https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_batch_normalization_grad_operator_desc

Copy pasta from the non-grad version. Broadcasting does *not* automatically happen for size 1 dimensions. Pass 0 for stride.


PRMerger15:

stevewhims (approved):

PR #1266, opened 2022-08-06
256-byte array => 256-bit array

radj307:
`BYTE` is 1 byte, not 256 bytes.

(`BYTE` == `unsigned char`)


PRMerger12:
@radj307 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1267, opened 2022-08-08
Update nf-pdh-pdhexpandwildcardpatha.md

ChrisOfTheLewi:
The PDH_REFRESHCOUNTERS flag isn't documented here.  It should be added as it is needed to get accurate information back about transient counters / counters added since the original wildcard counter was added and the time this was called.  This should also be edited in the different PdhExpandWildCardPath* functions which use the same flags.


PRMerger13:
@ChrisOfTheLewi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1270, opened 2022-08-13
The function is not exported from kernel32.dll

NN---:
https://stackoverflow.com/questions/47448559/getstagedpackageorigin-is-not-found-in-kernel32-dll-as-documented-but-in-kernel

https://stackoverflow.com/questions/54234075/lnk2019-unable-to-use-getstagedpackageorigin


drewbatgit:
Thanks for catching this!

PR #1275, opened 2022-08-18
Fix erroneous statement in CreateHardLinkA docs

Nisto:
Also see the following Stack Overflow question:
https://stackoverflow.com/questions/70119403/hard-link-to-a-symbolic-link-with-win32-api


GrantMeStrength:
Thank you for this correction. - John

PR #1277, opened 2022-08-22
RegDeleteTree: RegDeleteKeyTransacted docs say HKEY can be transacted here

oold:
Quote from `RegDeleteKeyTransacted`'s docs:
> If the function succeeds, RegDeleteKeyTransacted removes the specified key from the registry. The entire key, including all of its values, is removed. To remove the entire tree as a transacted operation, use the [RegDeleteTree](https://docs.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-regdeletetreea) function with a handle returned from [RegCreateKeyTransacted](https://docs.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-regcreatekeytransacteda) or [RegOpenKeyTransacted](https://docs.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-regopenkeytransacteda).


QuinnRadich:
#sign-off

PR #1280, opened 2022-08-26
Update nf-winuser-postmessagea.md

scottsmith-im:
Removed the following from the end of the "returns" section -- I think it's copy-pasta from somewhere else...

>  <b>GetLastError</b> returns <b>ERROR_NOT_ENOUGH_QUOTA</b> when the limit is hit.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-postmessagea.md
+++ b/sdk-api-src/content/winuser/nf-winuser-postmessagea.md
@@ -132,7 +132,7 @@ Type: <b>BOOL</b>
 
 If the function succeeds, the return value is nonzero.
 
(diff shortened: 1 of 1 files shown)


cts-davean:
Rather than removing the comment about ERROR_NOT_ENOUGH_QUOTA, it should instead be moved to the Remarks section of the article. GetLastError returns ERROR_NOT_ENOUGH_QUOTA when PostMessage fails due to exceeding the posted message limit described in the Remarks section.

jwmsft (approved):

jwmsft:
Accepting this request to remove info from the Returns section. I've submitted a separate PR to add the info to the remarks as suggested by @cts-davean.

PR #1295, opened 2022-09-21
Add types to GlobalUnlock

Kreijstal:


GrantMeStrength:
Thanks for this PR! We don't normally repeat the types in the headings for the parameters. Looking at other topics, it's not part of the standard styling. If I've read this wrong, feel free to re-open. -John

PR #1296, opened 2022-09-21
Add GlobalAlloc types

Kreijstal:


GrantMeStrength:
Thanks for this PR! We don't normally repeat the types in the headings for the parameters. Looking at other topics, it's not part of the standard styling. If I've read this wrong, feel free to re-open. -John

Kreijstal:
> -John

Oh.. okay, interesting because I thought it was common in winuser, for example
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messagebox
almost all the functions, enums, structs there have a type system in the documentation..

PR #1301, opened 2022-09-22
Update nn-dskquota-idiskquotacontrol.md

s77rt:
IDiskQuotaControl inherits from IConnectionPointContainer


alvinashcraft (approved):
Looks good. Thanks for the PR!

PR #1303, opened 2022-09-22
Update nf-ntsecapi-lsacallauthenticationpackage.md

brunovandooren:
Changed the description of the 'ProtocolStatus' parameter. This parameter doesn't receive a pointer to a status. 
The caller must always supply a pointer to a variable, and the ProtocolStatus will contain an NTSTATUS value after a successful call.


alvinashcraft (approved):
This change looks good. Thank you for updating our docs!

PR #1304, opened 2022-09-22
Update nf-dskquota-idiskquotacontrol-initialize.md

s77rt:
i think the correct error is `ERROR_NOT_SUPPORTED`
There is not such error `ERROR_NOTSUPPORTED`


alvinashcraft (approved):
Good catch. Thanks for your contribution!

PR #1311, opened 2022-09-28
Attempt fix of broken link to QUEUE_USER_APC_FLAGS

BKreisel:
Link to QUEUE_USER_APC_FLAGS from QueueUserApc2 is broken.  The markdown for the link on that page appears correct. Only discrepancy I could find was that every other page in processthreadsapi seemed to have a tech.root value of backup.


Karl-Bridge-Microsoft:
Thanks @BKreisel - this fix required a bit more work in the back-end and across a large number of docs. I made those updates in a private PR that should be live soon. Closing this PR.

BKreisel:
Awesome. Thanks for looking into it @Karl-Bridge-Microsoft !

PR #1315, opened 2022-10-04
changed RpcRT4 to Rpcrt4

badasahog:
changed RpcRT4 to Rpcrt4 to be more consistent with other docs and the file name itself


stevewhims:
I changed another 68 instances of this.

PR #1319, opened 2022-10-06
Update nf-dskquota-idiskquotauser-setquotalimit.md

s77rt:
changed "The default quota limit" to "The quota limit"
This has nothing to do with the default value, it's just confusing


alvinashcraft (approved):
Looks better. Thanks for clarifying this sentence.

PR #1323, opened 2022-10-12
I suppose  lpbi is lpbmi ?

MBaesken:
the docu mentions an lpbi parameter. But we do not have such a parameter but instead [in, out] lpbmi  , probably this is meant?


GrantMeStrength:
Thanks good catch!

PR #1357, opened 2022-11-11
Update nf-wingdi-createbitmap.md

DJm00n:
Fix long standing lpBits stride doc issue in CreateBitmap function
Add stride calculation formula from [BITMAPINFOHEADER structure docs](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader#calculating-surface-stride).
Same formula was in old [“DIBs and Their Use” article](https://learn.microsoft.com/previous-versions/ms969901(v=msdn.10)) from 1992.

Looks like **WORD** to **DWORD** alignment requirement was changed somewhere around Windows 3.0 when processor architecture was changed from 16bit to 32bit.

### DIFF
--- a/sdk-api-src/content/wingdi/nf-wingdi-createbitmap.md
+++ b/sdk-api-src/content/wingdi/nf-wingdi-createbitmap.md
@@ -50,9 +50,6 @@ api_name:
  - CreateBitmap
 ---
 
-# CreateBitmap function
-
-
 ## -description
 
 The <b>CreateBitmap</b> function creates a bitmap with the specified width, height, and color format (color planes and bits-per-pixel).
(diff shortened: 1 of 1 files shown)


DJm00n:
I may be wrong and CreateBitmap() is a legacy 16bit API that inherited 16bit/**WORD** alignment requirement. Or there is <8 bpp сondition for **WORD** vs **DWORD** alignment.
In any case - it should be documented properly.
@oldnewthing can you please elaborate?

stevewhims:
Thanks, I've forwarded to product team for their review.

-Steve

oldnewthing:
My quick check suggests that the existing text is correct. The bitmap data for CreateBitmap is still WORD aligned, not DWORD.

X24MS:
> My quick check suggests that the existing text is correct. The bitmap data for CreateBitmap is still WORD aligned, not DWORD.

I think so too.  I think we can share this check to make the world an easier place to navigate -

### -param lpBits [in]

A pointer to an array of color data used to set the colors in a rectangle of pixels. Each scan line in the rectangle must be word aligned (scan lines that are not word aligned must be padded with zeros).  The buffer size expected, cj, can be calculated using the formla:

    cj = (((nWidth*nPlanes*nBitCount + 15) >> 4) << 1) * nHeight;
        
If this parameter is <b>NULL</b>, the contents of the new bitmap is undefined.

stevewhims:
Thanks, all! I've updated the PR with the suggested content, and merged.

-Steve

DJm00n:
Thank you all for clarifying! This buffer size calculation was really annoying.

PR #1364, opened 2022-11-29
Update ns-d3d12-d3d12_buffer_rtv.md

BluBloos:
local testing demonstrates that `FirstElement` refers to the number of elements, not the number of bytes.


BluBloos:
I'd also want to add that it appears 

image

`WIN32_FIND_DATAA` is using the W version struct def.

stevewhims (approved):

PR #1370, opened 2022-12-06
Update ns-schannel-schannel_cred.md

jess-krynitsky:
SCHANNEL_CRED should be displayed as deprecated and replaced with SCH_CREDENTIALS in a banner at the top of this page. There was a note added but it did not display on the public site. Please advise if there is a better format.


jess-krynitsky:
@Andrei-Popov FYI

stevewhims (approved):

PR #1380, opened 2022-12-14
Add missing return values to MsiDatabaseOpenViewA.  The doc is still unclear whether a PMSIHANDLE can be specified for param phView.

Mike4Online:


Mike4Online:
## Notes:

I neglected to make a corresponding edit to the[ MsiDatabaseOpenViewW function documentation page](https://learn.microsoft.com/en-us/windows/win32/api/msiquery/nf-msiquery-msidatabaseopenvieww).  Please commit a corresponding change to that page as well.

## Related API documentation issues which I lack the knowledge to confirm:

### ISSUE 1:
In the **Remarks** section, both pages state:

> it is recommended to use variables of type PMSIHANDLE because the installer closes PMSIHANDLE objects as they go out of scope, whereas you must close MSIHANDLE objects

However, in the **Syntax** section, the type of parameter phView is specified as MSIHANDLE.  There is no explicit mention within the **Syntax** section that phView can be an MSIHANDLE or a PMSIHANDLE.

### ISSUE 2:
The result set for an executed view must be closed by [MsiViewClose](https://learn.microsoft.com/en-us/windows/win32/api/msiquery/nf-msiquery-msiviewclose).  This would happen prior to closing the MSIHANDLE for the view.  But if the view is initially opened using a PMSIHANDLE, there is no mention in [the documentation for MsiViewClose](https://learn.microsoft.com/en-us/windows/win32/api/msiquery/nf-msiquery-msiviewclose) whether MsiViewClose can accept a PMSIHANDLE.  The class PMSIHANDLE in Msi.h defines an implicit conversion operator that returns the raw MSIHANDLE, so passing a PMSIHANDLE to MsiViewClose should be fine.  But the SDK docs should explicitly indicate this is so.

drewbatgit:
@Mike4Online I'm accepting your change. Thanks for catching and fixing this error in the docs. I looked at the header files and it looks like PMSIHANDLE implements a casting operator to MSIHANDLE, so it should work for all APIs that accept MSIHANDLE. I've added a sentence about this behavior to the [Use PMSIHANDLE instead of HANDLE](https://learn.microsoft.com/en-us/windows/win32/msi/windows-installer-best-practices?branch=pr-en-us-1461#use-pmsihandle-instead-of-handle) section that is linked in the docs. The update should be live within ~24 hours.

Mike4Online:
> I've added a sentence about this behavior to the [Use PMSIHANDLE instead of HANDLE](https://learn.microsoft.com/en-us/windows/win32/msi/windows-installer-best-practices?branch=pr-en-us-1461#use-pmsihandle-instead-of-handle) section that is linked in the docs.

Thanks for doing this.

One other consideration that is not obvious is whether a PMSIHANDLE can be _assigned_ by other MSI functions which assign handles to MSIHANDLE objects.  If so, then a PMSIHANDLE can truly be used in place of an MSIHANDLE object throughout the SDK.  But if not this behavior should be noted.  Either way its an additional documentation update.

Mike4Online:
One last oddity: many of the MSI functions will return values such as ERROR_SUCCESS.  These values are typically long values, but the return type in most MSI functions is UINT.  This can involve a need to cast return values from long to UNIT.  But I have not found any mention of this in the SDK docs, or whether a return value could ever NOT be successfully cast to UINT without losing data.

PR #1384, opened 2022-12-21
Fix description of ranges param in IDODownload::Start

shishirb-MSFT:


GrantMeStrength:
Thanks for the update!

PR #1387, opened 2022-12-25
Update nf-winuser-enumdesktopsa.md

CitrusIce:
original doc says that if the first paramater is null, this function use current window station. but in fact if the first parameter is null, the function will return all window station name through callback function


GrantMeStrength:

PR #1391, opened 2023-01-01
Update nf-wingdi-wglcreatecontext.md

dkardel01:
Changed it parameter name to meaning full name


stevewhims (approved):

stevewhims:
Changing that content would have broken the topic (that is, prevented the description string from being rendered). That's because our build system uses the content of the header file to generate a unique identifier for the parameter (in this case, the parm was anonymous so the system generated an id). If we change the id in the content so that it no longer matches the id that the build system uses, then there's a disconnect and the description is lost. So the best we can do in the docs is the change I made here.

Thanks!
-Steve

stevewhims (approved):

PR #1392, opened 2023-01-01
Remove incorrect CreateProcess*() documentation regarding CREATE_UNICODE_ENVIRONMENT

tahonermann:
The documentation for the process creation APIs listed below previously stated the following with respect to the `lpEnvironment` parameter.
>   "If this parameter is NULL and the environment block of the parent process
>    contains Unicode characters, you must also ensure that dwCreationFlags
>    includes CREATE_UNICODE_ENVIRONMENT."
This requirement is not correct; at least not for any current versions of Windows. Note that the documentation does not state how a caller of these APIs would determine whether or not the parent process has a Unicode environment block and methods to do so are not obvious. In any case, the process creation APIs are as capable at making such a determination as the author of any code that calls these APIs would be.

The affected process creation APIs are:
- CreateProcessA()
- CreateProcessW()
- CreateProcessAsUserA()
- CreateProcessAsUserW()
- CreateProcessWithLogonW()
- CreateProcessWithTokenW()

This incorrect documentation has been a source of confusion for many years as is exhibited by discussion in the following forums.
- https://stackoverflow.com/questions/4206190/win32-createprocess-when-is-create-unicode-environment-really-needed
- https://microsoft.public.win32.programmer.kernel.narkive.com/J9U2SRdX/createprocess-and-environment

### DIFF
--- a/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-createprocessa.md
+++ b/sdk-api-src/content/processthreadsapi/nf-processthreadsapi-createprocessa.md
@@ -167,7 +167,7 @@ An environment block consists of a null-terminated block of null-terminated stri
 
 Because the equal sign is used as a separator, it must not be used in the name of an environment variable.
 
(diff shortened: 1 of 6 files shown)


tahonermann:

Karl-Bridge-Microsoft (approved):

PR #1396, opened 2023-01-06
Revise MsiFormatRecord regarding passing an empty string under /permissive-

Mike4Online:


Mike4Online (commented):

Mike4Online:

Your privacy is important to us. Read our Privacy Policy here: https://www.nuix.com/privacy-policy. This email may contain confidential or privileged information. If you believe you have received it in error, please notify the sender immediately and delete this message without copying or disclosing it.

Mike4Online:
My text edits offer a solution that may be valid for C++11; however, there may be a more general and better approach to take when compiling with `/permissive-` on older versions of VC++ or when writing a C11 (or older C standard) application.

So I'm sure something in the text should be changed to support compiling with `/permissive-`, but I do not know the best way to express what I am getting at.  Perhaps passing `(LPSTR)""` is a better approach (but I don't know if that is valid either).

So feel free to accept this PR if you confirm that my text suggestions are valid.  But consider rejecting this PR _in favor of a clearer alternative_ if you can come up with one.

Mike4Online:
I do not know what I need to do, if anything, to have this pull request merged.  It is labeled `do-not-merge` only because my forked repo lacked the yaml files to agree to Microsoft's policies, which I have since added.

I am not authorized to merge into the `MicrosoftDocs/sdk-api` repo.  I assume someone from Microsoft will need to do this.

drewbatgit:
Thank you for proposing this update to the docs. I consulted the feature owners for this area and they have made the determination that your suggested change is likely correct, but in their opinion overly-specific to be included in this page.  I apologize for rejecting this PR after the time you invested in it, but I must defer to the feature owners for these decisions. Thanks again for trying to help improve the documentation!

PR #1411, opened 2023-01-25
Update nf-winuser-drawiconex.md

DJm00n:
Add more information about how the icons are drawn.
https://devblogs.microsoft.com/oldnewthing/20101021-00/?p=12483

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-drawiconex.md
+++ b/sdk-api-src/content/winuser/nf-winuser-drawiconex.md
@@ -150,9 +150,7 @@ Draws the icon or cursor using the width and height specified by the system metr
 </dl>
 </td>
 <td width="60%">
-Draws the icon or cursor using the image.
-
-If <b>DI_MASK</b> flag is not set then the color bitmap (or XOR mask) of <b>HICON</b> is drawn with the <b>SRCCOPY</b> <a href="/windows/win32/api/wingdi/nf-wingdi-bitblt">raster operation code</a>.
+Draws the icon or cursor using the image. See Remarks.
 
 </td>
 </tr>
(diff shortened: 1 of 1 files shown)


DJm00n:
@oldnewthing can you please give some comments?

oldnewthing:
Instead of "color bitmap (or XOR mask)", I would say just "image bitmap" or just "image". Because "(or XOR mask)" is confusing; it sounds like the operation is an XOR, and why are you getting the image from the mask anyway?

Also needs some grammatical fine-tuning, and the DI_MASK and DI_IMAGE documentation doesn't cover the case where both flags are set. 

I think it would be better to defer discussion to Remarks, since the flags interact in complex ways.

> * DI_IMAGE
> Draws the icon or cursor using the image. See <b>Remarks</b>.
> * DI_MASK
> Draws the icon or cursor using the mask. See <b>Remarks</b>.
> * DI_NORMAL
> Combination of <b>DI_IMAGE</b> and <b>DI_MASK</b>. See <b>Remarks</b>.
>
> **Remarks**
> If only one of the <b>DI_IMAGE</b> and <b>DI_MASK</b> flags is set, then the corresponding bitmap is drawn with the <b>SRCCOPY</b> <a href="/windows/win32/api/wingdi/nf-wingdi-bitblt">raster operation code</a>. 
>
> If both the <b>DI_IMAGE</b> and <b>DI_MASK</b> flags are set:
> * If the icon is a 32-bit alpha-blended icon, then the image is drawn with <b>AC_SRC_OVER</b> <a href="/windows/win32/api/wingdi/ns-wingdi-blendfunction">blend function</a> and the mask is ignored.
> * For all other icons, the mask is drawn with the <b>SRCAND</b> <a href="/windows/win32/api/wingdi/nf-wingdi-bitblt">raster operation code</a>, and the image is drawn with the <b>SRCCOPY</b> <a href="/windows/win32/api/wingdi/nf-wingdi-bitblt">raster operation code</a>

However, this may still be incomplete. I haven't investigated what happens when drawing into a metafile.

DJm00n:
@oldnewthing thanks a lot!

jwmsft (approved):

PR #1413, opened 2023-01-27
Fix max size formula for USN_RECORD_V2

bzoz:
The result of provided MaximumChangeJournalRecordSize formula is 4 bytes to low. For a 255 max component it gives 572 whereas such file would need a 576 byte long buffer. This is most likely a typo, adding a WCHAR instead of subtracting it fixes this.

### DIFF
--- a/sdk-api-src/content/winioctl/ns-winioctl-usn_record_v2.md
+++ b/sdk-api-src/content/winioctl/ns-winioctl-usn_record_v2.md
@@ -85,7 +85,7 @@ The size in bytes of any change
 <td>
 <pre>  MaximumChangeJournalRecordSize = 
       ( MaximumComponentLength * sizeof(WCHAR) 
-        + sizeof(USN_RECORD)   - sizeof(WCHAR) );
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):

sdk-api-src/content/winioctl/ns-winioctl-usn_record_v2.md
```diff
@@ -85,7 +85,7 @@ The size in bytes of any change
 <td>
 <pre>  MaximumChangeJournalRecordSize = 
       ( MaximumComponentLength * sizeof(WCHAR) 
-        + sizeof(USN_RECORD)   - sizeof(WCHAR) );
+        + sizeof(USN_RECORD)   + sizeof(WCHAR) );
```
kobykahane, line 88: I don't think the subtraction here was a typo. The `FileName` field in `USN_RECORD` is a 1-element array, so one `WCHAR` is already included in `sizeof(USN_RECORD)`. The subtraction compensates for that. In other words, the result is the same as `FIELD_OFFSET(USN_RECORD, FileName)`.

PR #1419, opened 2023-01-31
Document File Object Id IOCTLs properly

henke37:
It sucks to have IOCTLs without proper buffer types, lets fix the file object id ones.


henke37:
It's rather iffy if `FSCTL_SET_OBJECT_ID` allows specifying the id to use or not, so the text changes pertaining to it is a bit of guesswork. I'd like to have it and `FSCTL_SET_OBJECT_ID_EXTENDED` properly documented before this PR is merged.

I'd also like to investigate `FILE_ID_DESCRIPTOR` to see how it relates.

Karl-Bridge-Microsoft (approved):

PR #1425, opened 2023-02-02
Fix base interface for IClosable

Sergio0694:
This PR fixes the base interface for `IClosable`, which is `IInspectable`, not `IUnknown`:


GrantMeStrength:
Thanks!

PR #1432, opened 2023-02-08
Minor fixes for CryptAcquireContext and CatEnum...

ptorr-msft:
Clarified a confusing statement in `CryptAcquireContext`, and added extra notes about the return value of `CryptCATAdminEnumCatalogFromHash`


alvinashcraft (approved):
Looks good. Thanks for updating these topics, Peter.

PR #1455, opened 2023-03-02
Fix file encoding and remove pre-Vista comments

Robo210:
ETW changed drastically in Windows Vista, and comments about  pre-Vista behaviors are no longer helpful and may be adding confusion.

Also, the page had a bad character encoding after each instance of the word "Windows", which shows up on MSDN and in VSCode, but not viewing the file on Github. Saving the file fixed that.


idigdoug (approved):
Looks fine. I probably would have left the minimum required version tags unchanged, but this is probably ok.

Karl-Bridge-Microsoft (approved):

PR #1456, opened 2023-03-03
Add correct return type to IsEqualGuid

Loghorn:


stevewhims:
Thanks, @Loghorn! You're right: that return type does need fixing. But AFAIK, the change in this PR won't fix the syntax block. I've gone ahead and worked around the issue (in all 3 affected topics in this header) by overriding the syntax block. It's not ideal, but it's all we can do for now. My changes are live now, so I'll go ahead and close this PR.

Thanks!
-Steve

Loghorn:
It was a long shot because I've not found a document explaining how to change the syntax block, so I looked for other similar APIs and checked what was different 😄 
Thank you for fixing it! 👏

PR #1467, opened 2023-03-09
Update ns-wingdi-logfontw.md

DTBob:
Change incorrect text "Boolean OR" to correct text "bitwise OR (|)"


GrantMeStrength:
Makes sense, thanks!

PR #1472, opened 2023-03-13
Update ne-secext-extended_name_format.md

HerbertMauerer:
This still needs PG confirmation, esp. that it is not supported for AAD and MSA users. I did source research, but maybe the PG wants to start a DCR for this.

### DIFF
--- a/sdk-api-src/content/secext/ne-secext-extended_name_format.md
+++ b/sdk-api-src/content/secext/ne-secext-extended_name_format.md
@@ -99,7 +99,7 @@ The DNS domain name followed by a backward-slash and the SAM user name.
 
 ### -field NameGivenName:13
 
-The first name or given name of the user. Note: This type is inly available for GetUserNameEx calls for an Active Directory User.
+The first name or given name of the user. Note: This type is only available for **GetUserNameEx** calls for an Active Directory user.
 
 ### -field NameSurname:14
 
--- a/sdk-api-src/content/secext/ne-secext-extended_name_format.md
+++ b/sdk-api-src/content/secext/ne-secext-extended_name_format.md
@@ -103,7 +103,7 @@ The first name or given name of the user. Note: This type is only available for 
 
 ### -field NameSurname:14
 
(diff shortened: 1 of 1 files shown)


alvinashcraft:
@HerbertMauerer -  thanks for the contribution! I'm reaching out to some folks on the product side to see if we can confirm this information.

Also - you have "only" misspelled as "inly" in each Note, if you want to update that while we wait on confirmation.

sdk-api-src/content/secext/ne-secext-extended_name_format.md
```diff
@@ -99,8 +99,12 @@ The DNS domain name followed by a backward-slash and the SAM user name.
 
 ### -field NameGivenName:13
 
+The first name or given name of the user. Note: This type is inly available for GetUserNameEx calls for an Active Directory User.
```
alvinashcraft:
```suggestion
The first name or given name of the user. Note: This type is only available for **GetUserNameEx** calls for an Active Directory user.
```

sdk-api-src/content/secext/ne-secext-extended_name_format.md
```diff
@@ -99,8 +99,12 @@ The DNS domain name followed by a backward-slash and the SAM user name.
...
+
 ### -field NameSurname:14
 
+The last name or surname of the user. Note: This type is inly available for GetUserNameEx calls for an Active Directory User.
```
alvinashcraft:
```suggestion
The last name or surname of the user. Note: This type is only available for **GetUserNameEx** calls for an Active Directory user.
```

alvinashcraft (commented):
I think this makes the docs more accurate, given what has been discussed in the email thread shared with me today. I've added a couple of minor suggestions inline. We can come back and make additional changes later if the product team feels it would make things clearer.

alvinashcraft (approved):
Linda from the PG approves of the changes and they look good to me. Thanks again for your contribution!

PR #1475, opened 2023-03-17
Consulting the use of sdk-api: CreateProcessAsUser

httmatang:
hello, 
I encountered a problem when using the sdk-api of CreateProcessAsUser, which is described as follows:
I registered a service that runs on the session0. then I call the api of CreateProcessAsUser in the service to run powershell.exe as a user Identity, Probability error reporting, the error code is -1073741502(STATUS_DLL_INIT_FAILED)

Looking forward to reply. Thanks.


GrantMeStrength:
Thanks for your comment.

As this is not really a docs issue, the best place to get support is by asking on the Microsoft Q and A forum:

https://learn.microsoft.com/answers/questions/

I will close this issue now

-John

PR #1489, opened 2023-04-06
Add a VBA syntax example

JamieG-UK:
Because VBA cannot pass the pt argument ByVal using the POINTAPI structure, it has to pass it inline. There are a lot of incorrect examples online so clarifying this here would be great. This is the correct VBA declaration:

Declare PtrSafe Function MonitorFromPoint Lib "user32.dll" ( _
                                               ByVal X As Long, _
                                               ByVal Y As Long, _
                                               ByVal dwFlags As Long) As LongPtr

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-monitorfrompoint.md
+++ b/sdk-api-src/content/winuser/nf-winuser-monitorfrompoint.md
@@ -60,6 +60,13 @@ The <b>MonitorFromPoint</b> function retrieves a handle to the display monitor t
 
 ### -param pt [in]
 
+VBA Syntax
+
+Declare PtrSafe Function MonitorFromPoint Lib "user32.dll" ( _
+                                               ByVal X As Long, _
+                                               ByVal Y As Long, _
+                                               ByVal dwFlags As Long) As LongPtr
+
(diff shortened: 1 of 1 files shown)


GrantMeStrength (changes requested):
The VBA declaration is placed inside the `-param pt` parameter description, which breaks the structured parameter documentation. It should go in the `-remarks` section instead, and needs a proper fenced code block.

Also note: SDK API reference pages don't typically include VBA syntax — that's normally covered in language-specific wrapper libraries or community resources. If this is included, it should be framed as a community note in Remarks, e.g.:

```
## -remarks

**VBA note:** Because VBA cannot pass a `POINT` structure ByVal, declare the function with separate `X` and `Y` parameters:

```vb
Declare PtrSafe Function MonitorFromPoint Lib "user32.dll" ( _
    ByVal X As Long, _
    ByVal Y As Long, _
    ByVal dwFlags As Long) As LongPtr
```
```

GrantMeStrength:
@copilot /review

GrantMeStrength (changes requested):
**Review: Request Changes**

The VBA declaration is placed inside the `-param pt [in]` parameter description section. Function declarations should go in a dedicated Examples or Syntax section, not within a parameter description.

Additionally, VBA syntax is not typically included in Win32 API reference pages.

PR #1491, opened 2023-04-07
Incorrect return value and type

iPhilip:
The documentation for [this](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getchild) and the macros below display an incorrect return value and type. Those parts of the documentation are not accessible through this page so I added **(See Proposed Changes)** to the text in order to submit these proposed changes. Please ignore that string.

The type is stated as `void`. It should be `HTREEITEM`.

The Return value is stated as

> None

Based on [this](https://learn.microsoft.com/en-us/previous-versions/windows/embedded/aa453787(v=msdn.10)) parallel documentation page, it could/should be

> The handle to the item indicates success. **NULL** indicates failure.

The other macros that should be corrected in a similar way are:

1. [TreeView_GetDropHilight](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getdrophilight)
2. [TreeView_GetFirstVisible](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getfirstvisible)
3. [TreeView_GetLastVisible](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getlastvisible)
4. [TreeView_GetNextItem](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getnextitem)
5. [TreeView_GetNextSelected](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getnextselected)
6. [TreeView_GetNextSibling](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getnextsibling)
7. [TreeView_GetNextVisible](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getnextvisible)
8. [TreeView_GetPrevSibling](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getprevsibling)
9. [TreeView_GetPrevVisible](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getprevvisible)
10. [TreeView_GetSelection](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-treeview_getselection)

I hope the above makes sense and that this is the right way to submit these changes. I am still learning...

### DIFF
--- a/sdk-api-src/content/commctrl/nf-commctrl-treeview_getchild.md
+++ b/sdk-api-src/content/commctrl/nf-commctrl-treeview_getchild.md
@@ -45,7 +45,7 @@ api_name:
  - TreeView_GetChild
 ---
 
-# TreeView_GetChild macro
+# TreeView_GetChild macro (See Proposed Changes)
 
 
 ## -description


GrantMeStrength (changes requested):
The diff only adds "(See Proposed Changes)" to the page title — that text would appear verbatim on the published docs page and should not be merged. The actual return type/value fix you intended isn't present in the diff.

The return type for `TreeView_GetChild` (and the related macros) is auto-generated from the API metadata, so it may not be editable in this markdown file. If you can identify the specific text that's wrong and where it appears in the file, please update the PR with that change instead.

GrantMeStrength:
@copilot /review

GrantMeStrength (changes requested):
**Review: Request Changes**

This PR appends "(See Proposed Changes)" to the API title. This is not a valid documentation change — article titles must match the API name exactly.

Please revert the title change.

PR #1492, opened 2023-04-08
Correct definition of DXGI flip-model

zopsicle:
As mentioned in [1], the flip model also includes the FLIP_DISCARD swap effect.

[1]: https://devblogs.microsoft.com/directx/dxgi-flip-model/


zopsicle:
There are more places where `FLIP_DISCARD` is seemingly erroneously not mentioned:

 - https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiswapchain-present
 - https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect
 - https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/ne-dxgi1_2-dxgi_scaling
 - https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiswapchain-getbuffer

stevewhims (approved):

PR #1494, opened 2023-04-11
Update nf-winsock-bind.md parameters description

davidBar-On:
Third parameter is the length of the `addr` parameter, not `name` parameter.


stevewhims (approved):
The parameter name was correct; it was just the description that was wrong. I mean "correct" in the sense of "that's what's in the header". In the docs we need to use the name from the header in order for our authored description to render. Changing it as in the first commit would have caused that description to disappear entirely. :)

PR #1496, opened 2023-04-13
Change parameter value to avoid data type warning

Lipovlan:
The parameter `hProv` is of data type `HCRYPTPROV_LEGACY` which is `ULONG_PTR` which is `unsigned long long`. Thus when passing `NULL` as `hProv` to `CertOpenSystemStoreA` or `CertOpenSystemStoreW`, this will create a warning regarding incompatible pointer to integer conversion. Since `NULL` is just zero casted to `void *`, passing zero is an acceptable substitution.


Lipovlan:

alvinashcraft (approved):
This change looks good. Thank you for updating these values.

PR #1498, opened 2023-04-14
IAudioClient3::InitializeSharedAudioStream() returns AUDCLNT_E_INVALID_STREAM_FLAG if any flag other than AUDCLNT_STREAMFLAGS_EVENTCALLBACK is specified

davidebeatrici:
Related: https://social.msdn.microsoft.com/Forums/en-US/windowspro-audiodevelopment/thread/55b39a3d-0f2a-4f05-ac66-5c5abf32ec64


mvaneerde:
Thank you for removing the stream flags that are incorrectly documented as being valid, even though they are, as you noticed, actually invalid for this method.
It may also be worth adding that the [AUDCLNT_SESSIONFLAGS_XXX](https://learn.microsoft.com/en-us/windows/win32/coreaudio/audclnt-sessionflags-xxx-constants) values are also valid:
* `AUDCLNT_SESSIONFLAGS_EXPIREWHENUNOWNED`
* `AUDCLNT_SESSIONFLAGS_DISPLAY_HIDE`
* `AUDCLNT_SESSIONFLAGS_DISPLAY_HIDEWHENEXPIRED`

EDIT: never mind, I see on closer reading that the session flags are already documented as being supported. So this change is correct.

davidebeatrici:
No problem.

According to my testing, [`AUDCLNT_STREAMOPTIONS_MATCH_FORMAT` ](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/ne-audioclient-audclnt_streamoptions)([`IAudioClient2::SetClientProperties()`](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient2-setclientproperties)) is the proper replacement for `AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM` and presumably `AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY` as well.

As a bonus, [`IAudioClient::IsFormatSupported()`](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-isformatsupported) can be relied on as it takes the properties into account.

Please note that `AUDCLNT_E_UNSUPPORTED_FORMAT` is returned when asking for upmixing or downmixing.

PR #1501, opened 2023-04-17
Update nf-fileapi-getvolumeinformationa.md

SullyPanda76cl:
There was a bug in the value of the symbol FILE_SUPPORTS_GHOSTING, it says "0x00000200" but it should say "0x40000000"


alvinashcraft (approved):
Yes, you're correct. Thanks for catching this.

SullyPanda76cl:
this should be reflected on the learn.microsoft.com site?

alvinashcraft:
It should be, but there is currently a sync issue between the public and private sdk-api repos. I've notified the team and it should be resolved later today. Thanks again!

PR #1502, opened 2023-04-18
Several questions about AcquireNextFrame

HiEnLiang:
Hi Microsoft team，I used this demo to do a performance test on screen capture + encoding, and encountered several problems:
1. Assuming that the refresh rate of the display is set to 60, does the system forcibly refresh every 16.7ms, and then perform the capture operation? How is it achieved?

2. During the test, I found that AcquireNextFrame will have time jitter, sometimes exceeding 30ms. What are the factors that affect the jitter of acquireNextFrame? When a problem occurs, the CPU, memory, and io loads are not high.

### DIFF
--- a/sdk-api-src/content/dxgi1_2/nf-dxgi1_2-idxgioutputduplication-acquirenextframe.md
+++ b/sdk-api-src/content/dxgi1_2/nf-dxgi1_2-idxgioutputduplication-acquirenextframe.md
@@ -45,7 +45,10 @@ api_location:
 api_name:
  - IDXGIOutputDuplication.AcquireNextFrame
 ---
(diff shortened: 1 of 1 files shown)


stevewhims:
Hi, @HiEnLiang, and thanks for the questions. Since we don't want to merge these questions into the topic, and since it's not clear that changes/fixes need to be made to this documentation, I wondered if you'd mind visiting [Microsoft Q&A](https://docs.microsoft.com/answers/topics/uwp.html?sort=active&filter=all) to ask the questions there. That's going to be a god place to get answers for them. I'll close the PR.

Thank you! :)
-Steve

PR #1505, opened 2023-04-20
Update nf-fileapi-getvolumeinformationw.md

SullyPanda76cl:
another mistake in the symbol value


alvinashcraft (approved):
Thanks again for finding and updating this value.

PR #1506, opened 2023-04-20
Update nf-fileapi-getvolumeinformationa.md

SullyPanda76cl:
wrong flag value


alvinashcraft (approved):
Thanks again for the fix!

PR #1508, opened 2023-04-21
fix parameter doc error in getstreamingmode

j-leungyy:


drewbatgit (changes requested):
Can I recommend the following change:

Receives a value from the [MFSensorDeviceMode](ne-mfidl-mfsensordevicemode.md) enumeration specifying the streaming mode of the sensor process.

j-leungyy:
yes, updated wording.

---
In reply to: [1396497745](https://github.com/MicrosoftDocs/sdk-api/pull/1508#issuecomment-1396497745) [](http://example.com/codeflow?ancestors=1396497745,1396497745)

drewbatgit (approved):

PR #1517, opened 2023-05-07
Correct GetProcessInformation API

riverar:
See also: https://github.com/MicrosoftDocs/feedback/issues/1182


stevewhims:
There's some debate in https://github.com/MicrosoftDocs/feedback/issues/1182 as to whether this is the exact wording we want. So might be worth running by feature team folks.

-Steve

Karl-Bridge-Microsoft (approved):

PR #1523, opened 2023-05-15
Update ns-commctrl-nmtreevieww.md

iPhilip:
Changed link to NMHDR and TVITEM structures to reflect Unicode versions.


jwmsft (approved):

iPhilip:
I think the change to the change to the NMHDR link is incorrect. Please double-check me on that.

jwmsft:
Thanks for catching that. From what I can find, it should point to the richedit version, so the original link was correct. I've submitted a PR to revert that link.

PR #1524, opened 2023-05-15
Update ns-richedit-nmhdr.md

iPhilip:
Changed `idFrom` type to accurately reflect other references to the same member of the `NMHDR` structure. See [NMHDR structure (winuser.h)](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nmhdr).

### DIFF
--- a/sdk-api-src/content/richedit/ns-richedit-nmhdr.md
+++ b/sdk-api-src/content/richedit/ns-richedit-nmhdr.md
@@ -64,7 +64,7 @@ A window handle to the control sending the message.
 
 ### -field idFrom
 
-Type: <b><a href="/windows/desktop/WinProg/windows-data-types">UINT</a></b>
+Type: <b><a href="/windows/desktop/WinProg/windows-data-types">UINT_PTR</a></b>
 
 An identifier of the control sending the message.
 


Karl-Bridge-Microsoft:
Thanks for the suggestion! I see your point re. consistency, but the nmhdr struct in richedit.h is defined as:

typedef struct _nmhdr
{
	HWND	hwndFrom;
	UINT	idFrom;
	UINT	code;
} NMHDR;
#endif	// !WM_NOTIFY 

The docs have to reflect the actual source code definition, which is UINT.

iPhilip:
Thank you. I was confused about why they were defined differently. If you happen to know, I am curious which structures use the [winuser.h NMHDR structure](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nmhdr).

iPhilip:
I did some additional testing to verify the documentation. Please consider this code block:

```
#include <iostream> // for std::cout
#include <richedit.h>

int main()
{
	std::cout << sizeof(NMHDR) << " NMHDR" << std::endl;
	std::cout << "" << std::endl;

	INT_PTR ptr;

	NMHDR myNMHDR;
	ptr = (INT_PTR)&myNMHDR;
	std::cout << "NMHDR offsets:" << std::endl;
	std::cout << (INT_PTR)&myNMHDR.hwndFrom - ptr << std::endl;
	std::cout << (INT_PTR)&myNMHDR.idFrom - ptr << std::endl;
	std::cout << (INT_PTR)&myNMHDR.code - ptr << std::endl;
	std::cout << "" << std::endl;
	std::cout << "NMHDR sizes:" << std::endl;
	std::cout << sizeof(myNMHDR.hwndFrom) << std::endl;
	std::cout << sizeof(myNMHDR.idFrom) << std::endl;
	std::cout << sizeof(myNMHDR.code) << std::endl;
	std::cout << "" << std::endl;

	std::getchar();
	return 0;
}
```
The result for 64-bit is as follows:

```
24 NMHDR

NMHDR offsets:
0
8
16

NMHDR sizes:
8
8
4
```

The fact that the size of the second member of the structure is 8 bytes indicates that it cannot be of `UINT` type. It's likely to be of `UINT_PTR` type as in the [NMHDR structure (winuser.h)](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nmhdr).

PR #1525, opened 2023-05-15
Correct link text in GetClassWord alternative list

henke37:
Two separate links, two separate functions.


jwmsft (approved):

jwmsft:
Thanks for the correction.

PR #1530, opened 2023-05-17
Correct description for flag OFN_NOCHANGEDIR

ww-rm:
> Restores the current directory to its original value if the user changed the directory while searching for files.
This flag is ineffective for [GetOpenFileName](https://learn.microsoft.com/en-us/windows/desktop/api/commdlg/nf-commdlg-getopenfilenamea).

The original text states that `OFN_NOCHANGEDIR` is ineffective for [GetOpenFileName](https://learn.microsoft.com/zh-cn/windows/win32/api/commdlg/nf-commdlg-getopenfilenamea), but actually it works.

I'm not sure if it is indeed ineffective in some lower Windows versions, but it does work on Win10. Perhaps a more comprehensive description could be considered.

Affected Pages:

- OPENFILENAMEA - https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamea
- OPENFILENAMEW - https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew

### DIFF
--- a/sdk-api-src/content/commdlg/ns-commdlg-openfilenamea.md
+++ b/sdk-api-src/content/commdlg/ns-commdlg-openfilenamea.md
@@ -361,8 +361,6 @@ For old-style dialog boxes, this flag causes the dialog box to use long file nam
 <td width="60%">
 Restores the current directory to its original value if the user changed the directory while searching for files.
 
- This flag is ineffective for <a href="/windows/desktop/api/commdlg/nf-commdlg-getopenfilenamea">GetOpenFileName</a>.
-
 </td>
 </tr>
 <tr>


alvinashcraft:
Thanks for this PR, @ww-rm. I'm trying to track down somebody on the product team who can confirm the intended behavior and when it may have changed.

alvinashcraft:
I'm also seeing this comment on the enum value in some archived Team Foundation docs: https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/bb167321(v=vs.120)

PR #1537, opened 2023-05-17
Link directly to Window Class Styles page

henke37:
Cut out the middleman and directly link to the page with the list of styles.


jwmsft (approved):

jwmsft:
Thanks!

PR #1538, opened 2023-05-19
Update nf-ncrypt-ncryptdeletekey.md

ironyman:
Fix missing closing tag


alvinashcraft (approved):
Thanks for adding the missing closing tag for the div.

PR #1542, opened 2023-05-21
Update ns-winbase-file_name_info.md - Add error code `ERROR_MORE_DATA`

BinToss:
Until now, the documentation's Remarks section stated how to solve a particular error, but did not state which error code it would solve.


GrantMeStrength:
thank you!

PR #1545, opened 2023-05-24
Correct `WIN32_FIND_DATAA` struct

TedLyngmo:
The `WIN32_FIND_DATAA` `struct` was missing and instead the `WIN32_FIND_DATAW` `struct` had taken its place.

Also corrected two links to `GetShortPathNameA`.


GrantMeStrength:
Drew - you're the A vs W expert, want to take a look?

drewbatgit (approved):

drewbatgit:
@GrantMeStrength  Yes, It is not the programmatically generated syntax block that's being fixed here, it's a manual one that was just ported as-is from back when the A and W versions shared the same page. Approved. I'll complete the PR

PR #1548, opened 2023-05-30
Update ns-d3d12-d3d12_buffer_srv.md

amarpMSFT:
Clarify that structured buffer view stride must match shader struct size.


clandrew:
Looks great!

stevewhims (approved):

PR #1553, opened 2023-06-03
ns-verrsrc-vs_fixedfileinfo.md: dwFileDateMS/dwFileDateLS

skissane:
In practice, `dwFileDateMS` and `dwFileDateLS` are almost never used and are expected to be zero.

From what I can gather, Microsoft's own build tools have never set them to a non-zero value, going back to Windows 3.x. It appears very few third party tools do so either. While executables having non-zero values for this field reputedly exist, nobody seems to be able to produce an example of one.

And yet, the documentation doesn't mention this significant fact–I think it should.


GrantMeStrength:

PR #1555, opened 2023-06-05
Mark UWP API's allowed in Windows 10240

robUx4:
These API's are allowed in UWP since the first Windows 10 build.


GrantMeStrength:
Thanks!

PR #1569, opened 2023-06-20
Update nf-d2d1_3-id2d1devicecontext2-createimagesourcefromdxgi.md

adrianstephens:
I found that 3 x DXGI_FORMAT_R8 does not work (get an unhelpful invalid argument error), but by stepping through the disassembly discovered they need to be 3 x DXGI_FORMAT_A8. Can someone verify this?


stevewhims:
Thanks, @adrianstephens. This is possibly more of a GitHub issue than a PR since it needs investigation. In this case the investigation needs a good chunk of bandwidth to make sure we get the right answer. So we're going to put this on our backlog, and close the PR meanwhile. When we get around to it, we'll reopen and update this PR, and make any necessary doc changes.

Thanks again!
-Steve

adrianstephens:
Understood, and thanks. I had left feedback on the page but then I thought this would be a better way to me to track my findings. Didn't think of posting it as an issue!

PR #1570, opened 2023-06-21
Update nf-cfapi-cfupdateplaceholder.md

kenjiuno:
`the call with be failed with HRESULT(ERROR_CLOUD_FILE_...).` will be typo of
`the call will be failed with HRESULT(ERROR_CLOUD_FILE_...).`


alvinashcraft (approved):
Thanks for finding and fixing these typos!

PR #1588, opened 2023-07-05
Update nf-wintrust-winverifytrust.md

SBlack1974:
The text "The <b>EnableCertPaddingCheck</b> key will be set to "1" by default on June 10, 2014." was removed because this article: https://learn.microsoft.com/en-us/security-updates/SecurityAdvisories/2014/2915720
 states:

V1.4 (July 29, 2014): Revised advisory to announce that Microsoft no longer plans to enforce the stricter verification behavior as a default functionality on supported releases of Microsoft Windows. It remains available as an opt-in feature. See the Advisory FAQ section for more information.


alvinashcraft (approved):
Thanks for catching this. This change looks good to me.

PR #1592, opened 2023-07-12
Change all links to fd_set in nf-winsock2-select.md to point to the structure instead of the macro

KellanClark:


GrantMeStrength:
Thank you!

PR #1596, opened 2023-07-17
Update nf-processthreadsapi-initializeprocthreadattributelist.md

gtworek:
DeleteProcThreadAttributeList suggestion and link are not required, as this function does absolutely nothing (it's a simple `ret`).


Karl-Bridge-Microsoft:
Hi @gtworek - can you clarify? The remark states to call the DeleteProcThreadAttributeList function for cleanup. Not sure what the issue is with that guidance.

gtworek:
The remark says it's for cleanup, but this function does absolutely nothing, and there is nothing to clean up in this particular case. Of course it does not hurt if you call it, but at the same time we can make the documentation a bit better, and that's exactly my goal.

Karl-Bridge-Microsoft:
Thanks again, @gtworek.

Based on discussion with the kernel team, I'm going to reject this PR. The recommendation still stands. 

The fact that the Delete API is a no op is an implementation detail and doesn’t change the fact that people ought to call the API.

PR #1599, opened 2023-07-23
the pMediaType parameter doesn't exist, and GUID_NULL doesn't fit in a hresult

fsb4000:


drewbatgit:
@fsb4000 I was able to find the line of code in the source tree that returns this error:

if( NULL == m_pMediaType )
    {
        *pguidStreamType = GUID_NULL;
    }

Since you are correct that GUID_NULL doesn't fit in an HRESULT, it seems like this behavior is undefined. And since this is in archived content, I think the right thing to do is err on the side of caution and go ahead and remove this value from the table.

Alcaro:
No, that code is correct and not undefined. It's writing a GUID to a variable of type GUID.

But the target variable isn't the return value, unless you consider the out parameter to be the return value.

drewbatgit:
Ah, I see. The value is correct for the output parameter, but it was correct to remove this from the return values table. I will add the  information back in, but in the pguidStreamType parameter description.

PR #1612, opened 2023-08-21
Add C++/WinRT links to IInspectable member functions

citelao:
Today, the documentation IInspectable's member functions (`GetIids`, `GetRuntimeClassName`, and `GetTrustLevel`) does not mention their C++/WinRT equivalents. These wrappers are not obvious, since they are free functions in C++/WinRT and member functions in raw WinRT, _plus_ they are named differently in C++/WinRT.

This change adds simple links to the C++/WinRT helper.


stevewhims:
"See also" should be just a list of links, so I removed the annotations. If we want that kind of explanation, maybe put the link plus annotation in Remarks, or somewhere else in the body of the topic.

PR #1614, opened 2023-08-26
Document unexpected CheckColorSpaceSupport() behaviour

Filoppi:
CheckColorSpaceSupport() doesn't seem to ever return that scRGB is supported, even if the monitor is running in HDR and even if you can call SetColorSpace1() and set the swapchain to scRGB succesfully.

Maybe the documentation page should specify that it checks for the current color space of the monitor? Which of course isn't ever scRGB as of 2023.
In that case, the name of the function would be confusing.
Though maybe it's just an unintended bug?
Your own code samples seem to not have been affected by that issue:
https://github.com/microsoft/Windows-universal-samples/blob/main/Samples/D2DAdvancedColorImages/cpp/D2DAdvancedColorImages/DeviceResources.cpp#L329
but all my tests on multiple machines show that `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709` always returns as unsupported. This has been confirmed by multiple people.
Proof:
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/VideoBackends/D3DCommon/SwapChain.cpp#L166

### DIFF
--- a/sdk-api-src/content/dxgi1_4/nf-dxgi1_4-idxgiswapchain3-checkcolorspacesupport.md
+++ b/sdk-api-src/content/dxgi1_4/nf-dxgi1_4-idxgiswapchain3-checkcolorspacesupport.md
@@ -1,7 +1,7 @@
 ---
 UID: NF:dxgi1_4.IDXGISwapChain3.CheckColorSpaceSupport
 title: IDXGISwapChain3::CheckColorSpaceSupport (dxgi1_4.h)
-description: Checks the swap chain's support for color space. Some color spaces like `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709` (scRGB) might be returned as unsupported even if calling <a href="windows/win32/api/dxgi1_4/nf-dxgi1_4-idxgiswapchain3-setcolorspace1">SetColorSpace1</a> would succesfully set them, independently of whether the current display in running in Advanced Color mode or not.
+description: Checks the swap chain's support for color space.
 helpviewer_keywords: ["CheckColorSpaceSupport","CheckColorSpaceSupport method [DXGI]","CheckColorSpaceSupport method [DXGI]","IDXGISwapChain3 interface","IDXGISwapChain3 interface [DXGI]","CheckColorSpaceSupport method","IDXGISwapChain3.CheckColorSpaceSupport","IDXGISwapChain3::CheckColorSpaceSupport","direct3ddxgi.idxgiswapchain3_checkcolorspacesupport","dxgi1_4/IDXGISwapChain3::CheckColorSpaceSupport"]
 old-location: direct3ddxgi\idxgiswapchain3_checkcolorspacesupport.htm
 tech.root: direct3ddxgi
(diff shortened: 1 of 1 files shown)


Filoppi:
I made some tests regarding this, and it seems like the function is exclusively returning true if the swap chain is already in the color space that was queried, or if it's supported as a display output mode.

For example, querying for `DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020` always returns true on my system, because my monitor supports BT.2020. On SDR monitors, it returns false.
But querying for `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709` exclusively returns true if the swap chain had been changed to that color space through `SetColorSpace1()`, or if it had been created with `DXGI_FORMAT_R16G16B16A16_FLOAT` as `DXGI_SWAP_CHAIN_DESC::format`.

Another sample:
https://github.com/SpecialKO/SKIF/blob/c912d15f8a7c7bd683d81b9a5d7306f5bc3e5879/src/imgui/D3D11/imgui_impl_dx11.cpp#L1417

It seems like there's either a bug underneath, or that the documentation and naming of this function is a bit lacking :).
Thanks.

stevewhims (changes requested):
The new info in the description YAML field needs moving to the description section in the body of the topic, or into Remarks.

And the url needs a leading forward slash.

Filoppi (commented):
Solved

Filoppi:
@stevewhims I think I took care of the changes. Thanks.

stevewhims:
This change looks like it's documenting a bug (which we don't do). However, I'll take this PR, but please let's not document bugs in the docs. For bugs, there are channels that customers can pursue. :) Thanks!

stevewhims (approved):

Filoppi:
> This change looks like it's documenting a bug (which we don't do). However, I'll take this PR, but please let's not document bugs in the docs. For bugs, there are channels that customers can pursue. :) Thanks!

Thanks. Could you point me to where I could report a bug as this one?

stevewhims:
> > This change looks like it's documenting a bug (which we don't do). However, I'll take this PR, but please let's not document bugs in the docs. For bugs, there are channels that customers can pursue. :) Thanks!
> 
> Thanks. Could you point me to where I could report a bug as this one?

Yes, here are all of the developer support resources:

- [Microsoft Q&A](https://docs.microsoft.com/answers/topics/windows-uwp.html) is the right place to lobby for a change or a new feature, to make a suggestion, or to ask a product question.
- For community assistance with development questions, there are our [forums](https://social.msdn.microsoft.com/Forums/newthread?category=windowsapps&forum=wpdevelop&prof=required), and [StackOverflow](https://stackoverflow.com/questions).
- The [Windows developer support](https://developer.microsoft.com/windows/support) site explains other support options.
- The [Visual Studio developer community](https://developercommunity.visualstudio.com) is available, and responsive. The former User Voice feature is now there on the **Suggest a feature** tab, if you want to lobby for a change, a feature, or to make a suggestion.
- For developer issues, or suspected product bugs, there's [Microsoft Support](https://support.microsoft.com/contactus); a support engineer will look at the repro and see whether the issue is a Microsoft product bug, or something else. If it's a bug with a Microsoft product then the ticket will be free of charge.

Thanks!
-Steve

stevewhims:
My apologies, it was premature of me to take the changes in this PR. So I've reverted the change in https://github.com/MicrosoftDocs/sdk-api/pull/1621. The product team is currently investigating the behavior, and we won't be documenting that behavior at this time (it seems more likely that the behavior will want to be changed).

Thanks!
-Steve

Filoppi:
@stevewhims no problem, glad it's being investigated. Is there any way I could track the progress?

stevewhims:
> @stevewhims no problem, glad it's being investigated. Is there any way I could track the progress?

Let's use this thread to track progress. Even though it's merged (my bad), I'm keeping a link to it.

Filoppi:
Hey @stevewhims, are there any updates on this?

Filoppi:
> > @stevewhims no problem, glad it's being investigated. Is there any way I could track the progress?
> 
> Let's use this thread to track progress. Even though it's merged (my bad), I'm keeping a link to it.

All good :)?

Filoppi:
A new version of this was merged :)

PR #1618, opened 2023-09-03
Fix typo in nf-winnls-lcmapstringex.md

ericmutta:
In the "Return value" section, the first paragraph already talks about what is returned when the function is used for string mapping. The second paragraph intended to say what is returned when the function is used to generate a sort key, but repeated the words for the string mapping scenario.


Karl-Bridge-Microsoft (approved):

ericmutta:
Many thanks for following up @Karl-Bridge-Microsoft :+1:

PR #1622, opened 2023-09-07
SetCurrentDirectoryW does not support \\?\ prepended path that exceed MAX_PATH

cts-davean:
SetCurrentDirectoryW fails with ERROR_FILENAME_EXCED_RANGE when called with a path prepended with \\?\ and exceeds MAX_PATH. Confirmed this behavior on Windows 7 - Windows 11 with a test application and code inspection.

The current version of the documentation states that you can prepend the path with \\?\ to bypass the MAX_PATH limitation.


GrantMeStrength:
thanks!

PR #1633, opened 2023-09-23
Update nf-projectedfslib-prjstartvirtualizing.md

gtworek:
Initialization with PrjCommandCallbacksInit is not possible and not required.


GrantMeStrength (commented):
Thanks!

PR #1636, opened 2023-09-26
Improved AclSize description to mention the potential unused memory

gaogaotiantian:
There's no guarantee that `AclSize == sizeof(ACL) + getsize(ACEs)`. It's possible that there is allocated but unused memory in the ACL. We should mention that because people are making false assumptions on it.


sdk-api-src/content/winnt/ns-winnt-acl.md
```diff
@@ -66,7 +66,7 @@ Specifies a zero byte of <a href="/windows/desktop/SecGloss/p-gly">padding</a> t
 
 ### -field AclSize
 
-Specifies the size, in bytes, of the ACL. This value includes both the <b>ACL</b> structure and all the ACEs.
+Specifies the size, in bytes, of the ACL. This value includes the <b>ACL</b> structure, all the ACEs and the potential unused memory.
```
alvinashcraft:
```suggestion
Specifies the size, in bytes, of the ACL. This value includes the **ACL** structure, all the ACEs, and the potential unused memory.
```

I think it sounds OK with "the". I added a comma, and I also try to update any HTML formatting to markdown when I make updates to docs.

alvinashcraft (approved):

PR #1639, opened 2023-10-02
Fix incorrect links in STACKFRAME64 documentation

jeremy-rifkin:
A number of links on https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/ns-dbghelp-stackframe64 link to the 32-bit version of what they are talking about rather than the 64-bit version.


Karl-Bridge-Microsoft (approved):

jeremy-rifkin:
Thanks for the lightning fast response :)

PR #1640, opened 2023-10-03
[RemoveDirectoryW] clarify removes junction itself, target directory not affected

glenn-slayden:
The existing text has a couple confusions in it. Since the "state of the target" directory is essentially unrelated to the `RemoveDirectory` operation when the specified path refers to a directory junction, it was unclear what the point of the paragraph was at all. Namely, it was trying to point out that there is a special case for behavior that is detailed elsewhere on the page--but without clearly indicating exactly which behavior it was modifying.

So one was left wondering and worrying about why the target directory is being mentioned at all, for the case of deleting a junction.

My goal here is primarily to assure developers that, indeed, `RemoveDirectory` will *not*--and in fact, *never*--delete the target directory and its contents when removing a directory junction, The previous text seemed to raise the spectre of the opposite.

Please feel free to review, edit, or discard my attempt to remedy this.

### DIFF
--- a/sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
@@ -93,7 +93,7 @@ To recursively delete the files in a directory, use the
 
 <b>RemoveDirectory</b> can be used to remove a directory junction. Since the target directory and its contents 
     will remain accessible through its canonical path, the target directory itself is not affected by removing a 
-    junction which targets it. For this reason, when <b>lpPathName</b> refers to a directory junction, 
+    junction which targets it. For this reason, when <i>lpPathName</i> refers to a directory junction, 
     <b>RemoveDirectory</b> will remove the specified link regardless of whether the target directory is 
     empty or not. For more information on junctions, see 
     <a href="/windows/desktop/FileIO/hard-links-and-junctions">Hard Links and Junctions</a>.
--- a/sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
@@ -91,9 +91,11 @@ The <b>RemoveDirectory</b> function marks a directory for
 To recursively delete the files in a directory, use the 
     <a href="/windows/desktop/api/shellapi/nf-shellapi-shfileoperationa">SHFileOperation</a> function.
 
-<b>RemoveDirectory</b> removes a directory junction, even 
-    if the contents of the target are not empty; the function removes directory junctions regardless of the state of 
-    the target object. For more information on junctions, see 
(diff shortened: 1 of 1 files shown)


sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
```diff
@@ -91,9 +91,11 @@ The <b>RemoveDirectory</b> function marks a directory for
 To recursively delete the files in a directory, use the 
     <a href="/windows/desktop/api/shellapi/nf-shellapi-shfileoperationa">SHFileOperation</a> function.
 
-<b>RemoveDirectory</b> removes a directory junction, even 
-    if the contents of the target are not empty; the function removes directory junctions regardless of the state of 
-    the target object. For more information on junctions, see 
+<b>RemoveDirectory</b> can be used to remove a directory junction. Since the target directory and its contents 
+    will remain accessible through its canonical path, the target directory itself is not affected by removing a 
+    junction which targets it. For this reason, when <b>lpPathName</b> refers to a directory junction,
```
alvinashcraft:
```suggestion
    junction which targets it. For this reason, when <i>lpPathName</i> refers to a directory junction, 
```

Minor change to use italics for params

sdk-api-src/content/fileapi/nf-fileapi-removedirectoryw.md
```diff
@@ -91,9 +91,11 @@ The <b>RemoveDirectory</b> function marks a directory for
...
 
-<b>RemoveDirectory</b> removes a directory junction, even 
-    if the contents of the target are not empty; the function removes directory junctions regardless of the state of 
-    the target object. For more information on junctions, see 
+<b>RemoveDirectory</b> can be used to remove a directory junction. Since the target directory and its contents 
+    will remain accessible through its canonical path, the target directory itself is not affected by removing a 
+    junction which targets it. For this reason, when <b>lpPathName</b> refers to a directory junction, 
+    <b>RemoveDirectory</b> will remove the specified link regardless of whether the target directory is 
+    empty or not. For more information on junctions, see 
     <a href="/windows/desktop/FileIO/hard-links-and-junctions">Hard Links and Junctions</a>.
```
alvinashcraft:
```suggestion
    <a href="/windows/win32/FileIO/hard-links-and-junctions">Hard Links and Junctions</a>.
```
Update href to avoid a redirect (while we're in here making changes).

alvinashcraft (commented):
These changes look good. Thanks for making the docs clearer. I added a couple of minor suggestions to the section, but the wording looks good to me.

glenn-slayden:
Do these changes need to be propagated to `RemoveDirectoryA` as well? If so, sorry that I did not do that.

alvinashcraft (approved):
This looks good. Thanks for thinking about RemoveDirectoryA. I'll take care of making the same updates in that topic today.

PR #1658, opened 2023-10-25
Update nf-dwmapi-dwmflush.md

iPhilip:
It's unclear what

> until the next present

means in this case. It might need another word or two. [[Ref](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmflush)]


stevewhims (approved):

iPhilip:
Thank you for the edits. It might be useful to add a link to the Present method and its interface ([this?](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgiswapchain-present)).

PR #1666, opened 2023-10-31
Escape broken links for contoso.com

Saisang:
https://dev.azure.com/msft-skilling/Content/_workitems/edit/173987


stevewhims:
None of these changes are necessary (and some are undesirable). So I'll close the PR without taking the changes.

If you have only "contoso.com" (without quotes) in the content, then that doesn't render as a hyperlink (so there can't be a broken link). So no need to escape ANY case like that.

If you see L"<some string>" (or variants of that), then that's example C++ source code representing a wide character string. So if you put backticks inside the quote marks, then you're breaking that example (backticks were never intended to be part of the example string, and will render it syntactically incorrect). There's no need to escape uris or urls that are inside quotes.

I've confirmed that none of these examples are being rendered in the web page as hyperlinks. So there's nothing to change here.

PR #1669, opened 2023-11-01
Augment winsock `FD_SET` macro

Rageking8:
The page is on the `FD_SET` macro yet the description is identical to the similarly named `fd_set` structure. Fixed up the description, document the behavior with duplicate file descriptor and added parameter descriptions.

PR #1673, opened 2023-11-04
Fix excess right angle bracket in `FD_SET`

Rageking8:
Fix a typo I introduced in #1669. Combed through the whole repo and found no other similar occurrences.

### DIFF
--- a/sdk-api-src/content/winsock/nf-winsock-fd_set.md
+++ b/sdk-api-src/content/winsock/nf-winsock-fd_set.md
@@ -57,7 +57,7 @@ A descriptor identifying a socket which will be added to the set.
 
 ### -param set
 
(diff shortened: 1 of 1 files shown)


stevewhims (approved):

stevewhims:
I know that the docs are currently a mix of markdown and markup, but they should be all markdown. So whenever we touch a line, it might be good to make improvements as we go (I think fixing the entire topic is too much to ask). Anyway, so the first pass for this line here would be to change the HTML bold formatting to markdown bold formatting. But the next pass would involve adding a hyperlink to the structure, and in that case (for API ref content), we omit the bold (we bold type names when there's no hyperlink). I've made those changes in this case.

PR #1674, opened 2023-11-06
Create `FD_CLR` macro

Rageking8:


Rageking8:
Sorry about the excess double quotes typo, did not notice it after changing the markup to markdown (same for the other `FD_XXX` macro PRs, since I copy and pasted).

stevewhims:
Unfortunately, I'll have to close this PR without taking it, for the reasons below.

The system doesn't recognize this macro, so we can't document it like this (that is, with a formal topic for it, in this repo). Even if we took this PR, these topics wouldn't be recognized by the system (since the system doesn't recognize the macro), so these topics wouldn't be published by the system.

And I don't think it's vital to have full, detailed, formal topics for this macro. This macro is already mentioned in [Select, fd_set, and FD_XXX Macros](https://learn.microsoft.com/windows/win32/winsock/select-and-fd---2). So if we need a little more info about them, then that might be a good place.

Rageking8:
@stevewhims Thanks for the detailed explanation. I would like to know if all identifiers recognized by the system is already documented? One such example is things like `afunix.h` (does not seem to be documented yet, but AFAIK is shipped with the Windows SDK).

stevewhims:
> afunix.h

So, sometime an entire header is deliberately and explicitly excluded from the docs. `afunix.h` is an example of that. Sometimes individual APIs are excluded from the docs. But sometimes those things happen due to oversight, so we strive to fill gaps in the documentation when we can and when it makes sense. We want to focus our energies on documenting the right things (things that customers will get the most benefit from calling) in the most appropriate ways.

PR #1676, opened 2023-11-06
Create `FD_ISSET` macro

Rageking8:


stevewhims:
Unfortunately, I'll have to close this PR without taking it, for the reasons below.

The system doesn't recognize this macro, so we can't document it like this (that is, with a formal topic for it, in this repo). Even if we took this PR, these topics wouldn't be recognized by the system (since the system doesn't recognize the macro), so these topics wouldn't be published by the system.

And I don't think it's vital to have full, detailed, formal topics for this macro. This macro is already mentioned in [Select, fd_set, and FD_XXX Macros](https://learn.microsoft.com/windows/win32/winsock/select-and-fd---2). So if we need a little more info about them, then that might be a good place.

PR #1683, opened 2023-11-13
Update nf-rpcdce-uuidtostring.md

DJm00n:
https://devblogs.microsoft.com/oldnewthing/20231113-00/?p=109011


stevewhims (approved):
It might be worth using (or converting to) markdown whenever we touch a line. Bringing everything to markdown will take a long time, but every little helps.

PR #1686, opened 2023-11-17
Update nf-bcrypt-bcryptgenrandom.md

prabhuintel:
To use this function in Kernel mode, Cng.lib has to be added. This line was explained in the later part. So it good to highlight in Requrements.


alvinashcraft (approved):
Thanks for catching this.

PR #1693, opened 2023-11-24
Normalize Titles

FuzzySecurity:
I am pulling in the sdk and ddi in an automation process and these titles are not in line with the rest of the sdk and ddi repository. I already handle this in code now but it should be a small fix to normalize these.


FuzzySecurity:

stevewhims (changes requested):
Marking this way until I've done some investigation. I need to check some things before I'm sure we want to take this change.

Thanks!

FuzzySecurity:
Yea these are the only titles that have a different format, there is one more in the ddi I think but I haven't opened a pr there yet

stevewhims:
UPDATE: I think, in conclusion, that I'd prefer not to make these changes (at least not at the moment; and if we make them later, then I can do that any time). My initial concern was "Does removing the quotes cause the topic to not build; and therefore become unpublished?". I wondered whether the leading underscore mandated escaping with quotes (because certain characters in titles in other repos do require the value to be in quotes there). The topic stubs (which are internal-only artifacts) don't have the quotes, though. Anyway, I experimented, and sure enough removing the quotes from the topic I tested was fine.

But the mediaobj.h topics here aren't being built and published at the moment. So I'd like to figure out why, and resolve that, before making any further changes to the source there (such as removing the quotes). I might make that change temporarily while I test what's going on, but I don't want to make it for its own sake.

And then lastly, nearly all of these topics belong to "legacy" features that have been superceded by other technologies. So updating the docs in any way isn't a priority.

So, I'll go ahead and close this PR. It sounds like in your own code you've worked around the inconsistency here, so I hope that not taking the PR won't be too much of a hardship.

-Steve

PR #1694, opened 2023-11-25
Update nf-winbase-lstrcmpiw.md

gtworek:
Fixed parameter type, as the doc refers to wchar version only.


GrantMeStrength:
thanks!

PR #1709, opened 2023-12-08
Note that using `SetCurrentDirectory` with a path longer than `MAX_PATH` will break `CreateProcess`

ChrisDenton:
Using `SetCurrentDirectory` with `longPathAware` to set a current directory longer than the `MAX_PATH - 2` limit will cause `CreateProcess` to fail with `ERROR_INVALID_INPUT`.

This is a pretty big caveat for anyone using `SetCurrentDirectory` with long paths as it will appear to work initially but then prevents spawning any child process, with an error code that does not make it obvious where the problem originated.

### DIFF
--- a/sdk-api-src/content/winbase/nf-winbase-setcurrentdirectory.md
+++ b/sdk-api-src/content/winbase/nf-winbase-setcurrentdirectory.md
@@ -72,7 +72,7 @@ By default, the name is limited to **MAX_PATH** characters.
 > Starting with Windows 10, Version 1607, you can opt in to remove the **MAX_PATH** limitation. For details, see the *Maximum path length limitation* section of [Naming files, paths, and namespaces](/windows/win32/fileio/naming-a-file).
 
 > [!IMPORTANT]
-> Unless the caller overrides the current directory, setting a current directory longer than **MAX_PATH** causes [CreateProcessW](/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) to fail.
+> Setting a current directory longer than **MAX_PATH** causes [CreateProcessW](/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) to fail.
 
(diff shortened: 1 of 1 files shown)


riverar:
Tagging @stevewhims for visibility; very noteworthy side effect when using long path awareness.

sdk-api-src/content/winbase/nf-winbase-setcurrentdirectory.md
```diff
@@ -72,6 +72,8 @@ By default, the name is limited to MAX_PATH characters.
 
 > [!TIP]
 > Starting with Windows 10, Version 1607, you can opt-in to remove the MAX_PATH limitation. See the "Maximum Path Length Limitation" section of [Naming Files, Paths, and Namespaces](/windows/win32/fileio/naming-a-file) for details.
+>
+> Note that setting a current directory longer than the MAX_PATH limitation will cause <a href="/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessw">CreateProcess</a> to fail unless the caller overrides the current directory.
```
riverar: Optional suggestion: "overrides the current directory" is a bit ambiguous here. Should we just nix that and leave that as an exercise to the reader?

```suggestion
> Note that setting the current directory to a path longer than the MAX_PATH limitation will cause <a href="/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessw">CreateProcess</a> to fail.
```

riverar (approved):

sdk-api-src/content/winbase/nf-winbase-setcurrentdirectory.md
stevewhims: Yes, I've removed that clause. I didn't know what that meant. If it's important, and if someone can explain it (in a different PR or GitHub issue), then I'll be happy to add it.

ChrisDenton: I just mean that `CreateProcess` allows setting a current directory explicitly and that this will still work. I am however happy to let users infer this for themselves.

PR #1710, opened 2023-12-08
Update nf-bcrypt-bcryptcreatehash.md

k-nero:
Fixed wrong flag's name


alvinashcraft (approved):
Thanks for fixing this. Looks good!

PR #1711, opened 2023-12-09
Update ns-wincrypt-crypt_algorithm_identifier.md

iPhilip:
Deleted extra lines in [Members](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-crypt_algorithm_identifier#members) section


alvinashcraft (approved):
Looks good. Thanks for your contribution!

PR #1716, opened 2023-12-15
Update ActivityCoordinator policy template labels

daniel-ju:
The title/label of the condition tables should be the policy template name.


alvinashcraft (approved):
Thanks for the update. Looks good.

PR #1728, opened 2024-01-04
Update nf-winuser-mapvirtualkeyexw.md

DJm00n:
MAPVK_VK_TO_VSC does not support extended keys with 0xe0/0xe1 prefix


DJm00n:
@Karl-Bridge-Microsoft any objections?

Karl-Bridge-Microsoft (approved):

PR #1729, opened 2024-01-05
Update ns-wincrypt-crypt_algorithm_identifier.md

manison:
Fix copy/paste error for szOID_RFC3161_counterSign constant.


alvinashcraft (approved):
Nice catch. The new value you entered matches what I see in the header file. Thanks for your contribution!

PR #1730, opened 2024-01-08
ws2tcpip: Fix GetAddrInfoEx parameters

jay:
- Remove GetAddrInfoExW reference to lpCompletionRoutine parameter type LPWSAOVERLAPPED_COMPLETION_ROUTINE.

- Clarify that lpCompletionRoutine parameter type is always LPLOOKUPSERVICE_COMPLETION_ROUTINE regardless of Windows version.

- Change GetAddrInfoExW references of lpNameHandle to lpHandle, since the latter name is used in the function prototype.

Prior to this change the lpCompletionRoutine parameter was erroneously documented for GetAddrInfoExW as LPWSAOVERLAPPED_COMPLETION_ROUTINE in the parameter details. Further, both GetAddrInfoExA/W implied that the correct type, LPLOOKUPSERVICE_COMPLETION_ROUTINE, was for Windows 8 and Windows 2012. In fact all versions of Windows and SDKs (including Windows 11) that support these functions expect parameter type LPLOOKUPSERVICE_COMPLETION_ROUTINE.

Assisted-by: edmcln@users.noreply.github.com

Ref: https://github.com/curl/curl/issues/12630

Closes #xxxxx

---

**The related issues below should also be addressed but there is no way to file them as issues.**

#1. GetAddrInfoExW prototype uses the name `lpHandle` for the last parameter and GetAddrInfoExA prototype uses the name `lpNameHandle` for the last parameter. It may be better for consistency with GetAddrInfoExA doc to modify the GetAddrInfoExW function prototype in the SDK to use lpNameHandle rather than change all references of that in nf-ws2tcpip-getaddrinfoexw.md to lpHandle like I did. The function prototype is not in nf-ws2tcpip-getaddrinfoexw.md so I cannot make the change myself.

#2. While investigating LPWSAOVERLAPPED_COMPLETION_ROUTINE (the incorrect type of lpCompletionRoutine) it was determined that the minimum supported client/server Windows 10 Build 20348 requirement is wrong, and it is used in much earlier SDKs and Windows versions.

/cc @edmcln @pps83

### DIFF
--- a/sdk-api-src/content/ws2tcpip/nf-ws2tcpip-getaddrinfoexa.md
+++ b/sdk-api-src/content/ws2tcpip/nf-ws2tcpip-getaddrinfoexa.md
@@ -247,7 +247,7 @@ An optional pointer to a function to be invoked upon successful completion for a
 
 This parameter is only supported when the <b>UNICODE</b> or <b>_UNICODE</b> macro has been defined in the sources before calling the <b>GetAddrInfoEx</b> function.
 
-On Windows 8 and Windows Server 2012, if this parameter is specified, it must be a pointer to a function with the following signature:
+If this parameter is specified, it must be a pointer to a function with the following signature:
 
 
 ```cpp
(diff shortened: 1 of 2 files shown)


jay:
The bad reference to LPWSAOVERLAPPED_COMPLETION_ROUTINE came from 6dc0a1cc. Other files in that commit have the same change but they all look correct because the function prototypes for those functions and structs have at least one parameter type of LPWSAOVERLAPPED_COMPLETION_ROUTINE.

/cc @stevewhims

stevewhims (approved):
Thank you!

PR #1736, opened 2024-01-17
Indicate TreeView_SortChildren recurse flag is NOTIMPL

riverar:
After investigating [a reported `TreeView_SortChildren` issue](https://github.com/microsoft/win32metadata/issues/1823), it was determined that the default TreeView sorting callback (`comctl32!TV_SortCB`) receives the recurse flag value but doesn't do anything with it. It remains unimplemented in Common Controls v5 and v6.

### DIFF
--- a/sdk-api-src/content/commctrl/nf-commctrl-treeview_sortchildren.md
+++ b/sdk-api-src/content/commctrl/nf-commctrl-treeview_sortchildren.md
@@ -70,7 +70,7 @@ Handle to the parent item whose child items are to be sorted.
 
 Type: <b><a href="/windows/desktop/WinProg/windows-data-types">BOOL</a></b>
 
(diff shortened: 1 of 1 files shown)


jwmsft (approved):

gonutz:
I just took a look at your change. You say `Not implemented; see remarks` but the remarks were not updated and they do not mention this at all.

riverar:
@gonutz That was intentional. It already says:

> [...] You can use the TVM_SORTCHILDRENCB message to customize the ordering behavior.

Do you think we're missing something there?

gonutz:
The way I read the doc page now is that the function has a parameter `recurse` which makes it pretty obvious what it is for. Then I read it is not implemented. That is OK (Maybe I would have changed the name here to something like `reserved`, but the SDK header files would still have it be `recurse` so this might create extra confusion). The part that looks like something is missing is the `see remarks` part here. This makes me expect a reason or further explanation for why this is not impemented. Escpecially since there is now no obvious way to sort a tree recursively. Users have to implement this themselves. The remarks explain only how to sort in a way other than alphabetically, not how to do it recursively.

PR #1741, opened 2024-01-22
FIX:  ")X04000" -> "0x04000"

spdkils:
Trying to write a python call, and as I'm copy/paste flags, this sticks out as a clear mistake.

Hex doesn't include )X someone got their shift key stuck.


GrantMeStrength:
Thank you!

PR #1743, opened 2024-01-24
Patch 2

Whitecx:
The table under the idHook parameter details hook procedure types one can specify when using SetWindowsHookExA, including links to the relevant callback functions. 

The row for WH_CALLWNDPROC references [CallWindowProcA](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-callwindowproca) (which the docs specify to use for subclassing) instead of the associated callback function, CallWndProc.

CallWindowProc
> Use the CallWindowProc function for window subclassing

CallWndProc
> An application installs the hook procedure by specifying the [WH_CALLWNDPROC](https://msdn.microsoft.com/en-us/library/ms644959(v=vs.85)) hook type and a pointer to the hook procedure in a call to the [SetWindowsHookEx](https://msdn.microsoft.com/en-us/library/ms644990(v=vs.85)) function

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-setwindowshookexa.md
+++ b/sdk-api-src/content/winuser/nf-winuser-setwindowshookexa.md
@@ -83,7 +83,7 @@ The type of hook procedure to be installed. This parameter can be one of the fol
 </td>
 <td width="60%">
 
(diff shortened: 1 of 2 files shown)


Whitecx:

jwmsft (approved):

PR #1753, opened 2024-02-05
Update nf-ocidl-ispecifypropertypages-getpages.md

iPhilip:
Deleted text that, in my opinion, creates confusion. The "must be initialized and filled" part seems to imply that the caller is supposed to initialize and fill the structure.


stevewhims (approved):

stevewhims:
Yes, I guess the original writer must have been thinking that *from the caller's point of view", the structure "must be initialized and filled before returning". Which, as you say, is unclear. I've put the clause back in, but I've made it clear who's doing the work. :)

iPhilip:
Great! It's even clearer with your edits. Thank you.

PR #1754, opened 2024-02-06
Make description match function signature

georgthegreat:


riverar:
Looks like this is already fixed in public docs and no longer needed. https://learn.microsoft.com/windows/win32/api/projectedfslib/nf-projectedfslib-prjstopvirtualizing

PR #1766, opened 2024-02-10
Update nf-gdiplusgraphics-graphics-drawdriverstring.md

iPhilip:
Fixed bold formatting issue in the [Parameters](https://learn.microsoft.com/en-us/windows/win32/api/gdiplusgraphics/nf-gdiplusgraphics-graphics-drawdriverstring#parameters) section.


stevewhims (approved):

stevewhims:
Thanks! I suspect that should have been formatted italic rather than bold (it seems to be a variable rather than a type); and formatting should be in markdown wherever possible (although it's not easy to see where it's possible or otherwise from outside of the org). There are lots of other things that could use an edit in this topic. Anyway, I took the change for expediency. Thanks! :)

PR #1770, opened 2024-02-19
winsock: Fix WSAStartup parameter description

jay:
winsock: Fix WSAStartup parameter description

- Replace wVersionRequired with wVersionRequested.

Prior to this change wVersionRequired had a description of TBD and wVersionRequested had the correct description though was not rendered.

Both are different names of the same parameter depending on which SDK is used. The function prototype is extracted from an SDK which uses wVersionRequired but IMO "required" is the wrong name to use. It is properly noted as the "requested" version numerous times throughout the documentation.

This change makes the tradeoff that rather than be consistent with what name the renderer has extracted from the SDK (ie wVersionRequired), instead use the name that is most correct (ie wVersionRequested).

Recent SDKs (including the latest) and even an old Windows 7 SDK have wVersionRequested. I couldn't find which SDK has wVersionRequired.

A complete fix for this issue would be to get the renderer to extract the function prototype from the latest version of the SDK.

Ref: https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup

Closes #xxxxx

### DIFF
--- a/sdk-api-src/content/winsock/nf-winsock-wsastartup.md
+++ b/sdk-api-src/content/winsock/nf-winsock-wsastartup.md
@@ -47,17 +47,13 @@ api_name:
  - WSAStartup
 ---
 
-# WSAStartup function
-
-
 ## -description
 
-The 
-<b>WSAStartup</b> function initiates use of the Winsock DLL by a process.
+The <b>WSAStartup</b> function initiates use of the Winsock DLL by a process.
 
 ## -parameters
 
-### -param wVersionRequested [in]
+### -param wVersionRequired [in]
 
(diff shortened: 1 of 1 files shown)


jay:
I've done a 180 on this and updated the branch to replace the name wVersionRequired with wVersionRequested since the latter makes more sense.

stevewhims:
`wVersionRequired` is the name in the SDK, so it's the name that the publishing pipeline will expect. If we were to make that change to the name of the parm (in the markdown), then that parm's section in the file would be ignored, and that description would fail to render.

stevewhims (approved):

stevewhims (approved):

PR #1773, opened 2024-02-21
Winsock2 function documentation fixes

wesinator:
many of the longstanding winsock2 / ws2_32 function docs have incorrect info in the minimum client version field. Many of the functions that have been around since XP show a minimum client version of Vista in the docs. Some correctly show XP.

I also found the listing of Desktop | UWP to be somewhat confusing.
image

It seems like the min UWP platform (Windows 8.1) should be listed as the second item?

this PR fixes the min client ver listed for the most common functions, and ones that describe being supported by XP in the docs. (I confirmed these functions are present in XP ws2_32).

There may be other functions with incorrect minimum client version info, and more systematic updates and fixes may be good.

[NO_TRAIN]::

### DIFF
--- a/sdk-api-src/content/winsock2/nf-winsock2-wsaaddresstostringa.md
+++ b/sdk-api-src/content/winsock2/nf-winsock2-wsaaddresstostringa.md
@@ -11,7 +11,7 @@ ms.keywords: WSAAddressToString, WSAAddressToString function [Winsock], WSAAddre
 req.header: winsock2.h
 req.include-header: 
 req.target-type: Windows
-req.target-min-winverclnt: Windows 8.1, Windows Vista [desktop apps \| UWP apps]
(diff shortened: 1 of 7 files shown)


stevewhims:
I'll go ahead and file a request to see whether our publishing pipeline can be made to insert correct Min client/server info each time it builds. That'll be more efficient/correct than trying to do it manually.

stevewhims (approved):

stevewhims:
I think this change is an improvement (it's more accurate), so I've taken it. But I don't think that we should list two different versions. There can be only one min client. The desktop/UWP thing isn't saying anything about versions. It's just talking about what app type you can call the API from. Also, Win 8.1 pre-dated UWP. I feel that these particular values should say "Windows XP [desktop | UWP]." It's ok for now, but if you make any more changes along these lines, I'd suggest doing that.

Thanks! :)
-Steve

PR #1771, opened 2024-02-20
Update nf-wow64apiset-wow64revertwow64fsredirection.md

wesinator:
fix documentation of Wow64RevertWow64FsRedirection param

This should be OldValue [in] which is reflected in original version https://github.com/MicrosoftDocs/sdk-api/commit/c59c21b25257e780426831d5777ffcf1c89343f7#diff-e16bf57fbe7232482325e817f081d23d7e060c5f42470bf7eecf4c67bd8e6b08R71

https://learn.microsoft.com/windows/win32/api/wow64apiset/nf-wow64apiset-wow64disablewow64fsredirection

PR #1776, opened 2024-02-26
Update nf-wow64apiset-wow64revertwow64fsredirection.md

wesinator:
This PR updates #1771 to retain the literal argument variable name from `wow64apiset.h`.

[NO_TRAIN]::


stevewhims:
In a reasonable amount of time, I couldn't make sense of `This value is _sic_, defined in`. Are you saying that the value is actually the string "sic"? "Sic" usually means "Yes, what I just said, or quoted, is as intended even though it looks wrong or is wrong". I don't think that's the case. Rather than completely remove that sentence, I tweaked it to something more feasible. If that still needs an edit, please create another PR.

Thanks! :)
-Steve

stevewhims (approved):

stevewhims:
Also, we don't say things like "Note that X". We either just say "X", or we turn the sentence into a `> [!NOTE]`.

PR #1782, opened 2024-03-06
Update nf-wingdi-createenhmetafilew.md

danielworkman:
Parameter names did not match those given in the remarks section


stevewhims (approved):

stevewhims:
I've undone the changes, because we can't change the parameter headings there (that will prevent the descriptions from being rendered). Instead I've made changes in Remarks, and also made a few opportunistic ad-hoc tweaks to the formatting (etc.) in the topic.

stevewhims:
Hi, it looks like the CLA hasn't been signed yet (please see comment in the thread above). If you could sign that, then the button will appear on this end for me to complete the PR. Thanks! :)

danielworkman:

PR #1784, opened 2024-03-08
Update nf-ntdsapi-dsaddsidhistorya.md

HerbertMauerer:
make it extra clear the user identity needs to be admin in the source domain.


GrantMeStrength:
Thanks!

PR #1792, opened 2024-03-17
Warn .net users of SeWindowsHookExW about garbage-collection of callbacks

asklar:


sdk-api-src/content/winuser/nf-winuser-setwindowshookexw.md
```diff
@@ -300,6 +300,8 @@ An error may occur if the <i>hMod</i> parameter is <b>NULL</b> and the <i>dwThre
 
 Calling the [CallNextHookEx function](nf-winuser-callnexthookex.md) function to chain to the next hook procedure is optional, but it is highly recommended; otherwise, other applications that have installed hooks will not receive hook notifications and may behave incorrectly as a result. You should call <b>CallNextHookEx</b> unless you absolutely need to prevent the notification from being seen by other applications.
 
+In .net, you must ensure the callback is not moved around by the garbage collector (otherwise your app will crash with an ExecutionEngineException). One way to do this is by making the callback a static method of your class.
```
jonwis: What do the other "set this method as a callback" docs say about this? We should probably be consistent. Also there's other GC'd runtimes beyond .NET...

tim-weis: [`EnumWindows()`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindows) (as an example) doesn't say anything, and I feel that is sufficient. Repeating ground rules for interfacing with C APIs shouldn't proliferate the API documentation.

If you feel that a warning is due, at the very least, make sure a developer who's not in the target audience can quickly determine, at a glance, that this information isn't for them.

jwmsft (approved):

PR #1793, opened 2024-03-18
Update ns-schannel-secpkgcontext_connectioninfo.md

fran-ki:
Developers will need to verify that the value zero is ALG_TYPE_ANY which I based on the value in the header file.  Also, I'm assuming it means any key exchange algorithm.


alvinashcraft (approved):
Thanks for adding this. I verified that this is correct for 0.

PR #1796, opened 2024-03-26
fix NOTIFYICONDATAA typo for ANSI version (should be CHAR instead of TCHAR)

expikr:


riverar (approved):

drewbatgit (approved):

PR #1801, opened 2024-04-06
Updated nf-cryptuiapi-cryptuiwizdigitalsign.md to include how to get extended information when an error occurs

ramiabughazaleh:


tim-weis:
This proposes a change to the API's contract. Has this been verified to match its implementation and *intended* contract?

The changes to the "Return value" section are wrong either way: It should continue to say *"Returns nonzero on success, zero on failure."* Changing the success code to read `TRUE` will effectively make it impossible to ever evolve the API and stay compatible with shipped code.

ramiabughazaleh:
This is from the Windows 11 SDK 10.0.22621.0, on my machine at `C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\cryptuiapi.h`:

```
//+----------------------------------------------------------------------------
//
// CryptUIWizDigitalSign
//
// The wizard to digitally sign a document or a blob.
//
// If CRYPTUI_WIZ_NO_UI is set in dwFlags, no UI will be shown.  Otherwise,
// user will be prompted for input through a wizard.
//
// dwFlags              IN Required: See dwFlags values above.
//
// hwndParent           IN Optional: The parent window handle.
//
// pwszWizardTitle      IN Optional: The title of the wizard.
//
// pDigitalSignInfo     IN Required: The information about the signing process.
//
// ppSignContext        OUT Optional: The context pointer points to the signed 
//                                    blob.
//
//-----------------------------------------------------------------------------
_Success_(return == TRUE)
BOOL
WINAPI
CryptUIWizDigitalSign(
    _In_                DWORD                               dwFlags,
    _In_opt_            HWND                                hwndParent,
    _In_opt_            LPCWSTR                             pwszWizardTitle,
    _In_                PCCRYPTUI_WIZ_DIGITAL_SIGN_INFO     pDigitalSignInfo,
    _Outptr_opt_        PCCRYPTUI_WIZ_DIGITAL_SIGN_CONTEXT  *ppSignContext
    );
```

Notice the `BOOL` as the return type.

The published documentation is currently wrong.
This PR does not propose any changes to the API's contract.

tim-weis:
A `BOOL` can be *any* 32-bit value, not just `FALSE` and `TRUE`, and you'll find the "zero/nonzero" wording for just about any API that returns a `BOOL` value (e.g., [`ShowWindow()`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow), or the classic [`GetMessage()`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmessage) example). This is not an oversight.

> This PR does not propose any changes to the API's contract.

It does introduce the guarantee that it was to set the calling thread's last error code on failure. That is a departure from the currently documented contract that makes no such promises. This needs to be verified to be true and intended. Only the owner of the API can make that call.

ramiabughazaleh:
The wording in this PR is based on
https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certclosestore

If the API could return anything other than zero or one, I expect the header file to indicate the return type as a long for example.
Since the header file indicates a return type of BOOL, I expect the documentation to indicate the return type values as TRUE or FALSE.

riverar:
> If the API could return anything other than zero or one, I expect the header file to indicate the return type as a long for example.
> Since the header file indicates a return type of BOOL, I expect the documentation to indicate the return type values as TRUE or FALSE.

Check out [`PathIsDirectoryA`](https://learn.microsoft.com/windows/win32/api/shlwapi/nf-shlwapi-pathisdirectorya) for an idea of how BOOL expectations can mismatch with reality.

That said, casual disassembly of `cryptuiwizard.dll!CryptUIWizDigitalSign` does appear to confirm only return values of `0` and `1`. I would feel more comfortable if someone at Microsoft could confirm however.

alvinashcraft:
I found the product team that owns these APIs, and they confirmed that only true and false (0 and 1) are possible return values for this function. I'm going to approve and merge the PR. Thanks everyone for the discussion!

alvinashcraft (approved):

tim-weis:
@alvinashcraft Did the product team that owns the API also confirm that they are setting the calling thread's last error on *every* error return path? Because that's what the documentation promises now.

alvinashcraft:
They said "When FALSE is returned, LastError will be set."

PR #1802, opened 2024-04-09
Provide descirption for fields

LuridHound:
Current version of msdn docs describing **enum D3D12_PIPELINE_STATE_FLAGS** doesn't cover 2 flags.  However information about them can be found from the microsoft devblog : https://devblogs.microsoft.com/directx/agility-sdk-1-608-0/ .


stevewhims (approved):

LuridHound:
What have I done wrong?
Even though changes are done - visually page stays the same.
https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_pipeline_state_flags

LuridHound:
@stevewhims , do you have any idea what went wrong?
Other enum fields are mentioned in file 4 more time. However looks like this is metadata for microsoft docs engine (search/something else).

stevewhims:
It looks like the public/private repos haven't sync'd yet. I'll investigate.

PR #1804, opened 2024-04-11
Update nf-winbase-movefileexa.md

ericlaw1979:
Name variable correctly. When deleting a file, there is no "destination" filename.


GrantMeStrength:
thanks!

PR #1805, opened 2024-04-18
Add note from `MS15-097` to `CloseClipboard`?

rich-ayr:
Adds a security note for `MS15-097` to `CloseClipboard`.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-closeclipboard.md
+++ b/sdk-api-src/content/winuser/nf-winuser-closeclipboard.md
@@ -79,6 +79,10 @@ When the window has finished examining or changing the clipboard, close the clip
 
 Do not place an object on the clipboard after calling <b>CloseClipboard</b>. 
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


oberrich:
I'm not affiliated with MSFT and coincidentally stumbled across `MS15-097` recently. As far as I understand, this is a commonly overlooked security issue that hasn't been (or can't be?) patched at the kernel level. Maybe the person internally responsible for this can first check if this is still applicable.

I have also raised this issue across clipboard libraries and am trying to learn more about this, see: [[x-issue] Potential EoP in Windows clipboard libraries (MS15-097)](https://github.com/oberrich/x-issues/issues/1).

rich-ayr:
I did some more research on this and it's definitely fixed on newer Windows kernels (I cba to find the kb number but it must've been some time between 2015-2022)

image
_Disassembly of `win32kfull!xxxCloseClipboard` (called by `win32kfull!NtUserCloseClipboard`, `...!...`, `user32!CloseClipboard`_)

This only affects older systems that haven't installed the relevant update (update was cumulative on some versions).
Given this information, I think this can be closed.

PR #1807, opened 2024-04-20
Clarify IContextMenu::QueryContextMenu return value

whindsaks:
It is very unhelpful to use 5 for both idCmdFirst and the menu id in the example!

I don't guarantee that my understanding of this calculation is correct. Please ask Raymond Chen to verify if the -5 in the example is supposed to be idCmdFirst or the lowest menu id.

### DIFF
--- a/sdk-api-src/content/shobjidl_core/nf-shobjidl_core-icontextmenu-querycontextmenu.md
+++ b/sdk-api-src/content/shobjidl_core/nf-shobjidl_core-icontextmenu-querycontextmenu.md
@@ -198,7 +198,7 @@ This value is not available.
 
 Type: <b>HRESULT</b>
 
-If successful, returns an <b>HRESULT</b> value that has its severity value set to SEVERITY_SUCCESS and its code value set to the offset of the largest command identifier that was assigned, plus one. For example, if <i>idCmdFirst</i> is set to 20 and you add three items to the menu with command identifiers of 5, 7, and 8, the return value should be MAKE_HRESULT(SEVERITY_SUCCESS, 0, 8 - 5 + 1). Otherwise, it returns a COM error value.
+If successful, returns an <b>HRESULT</b> value that has its severity value set to SEVERITY_SUCCESS and its code value set to the offset of the largest command identifier that was assigned, plus one. For example, if <i>idCmdFirst</i> is set to 5 and you add three items to the menu with command identifiers of 4, 7, and 8, the return value should be MAKE_HRESULT(SEVERITY_SUCCESS, 0, 8 - 5 + 1). Otherwise, it returns a COM error value.
 
 ## -remarks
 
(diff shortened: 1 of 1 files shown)


whindsaks:
There should probably also be a note somewhere that some shell extensions just return a count of items it added.

GrantMeStrength (changes requested):
The added parenthetical "(zero based identifiers + idCmdFirst)" is misleading. In the example, idCmdFirst is 5 and the assigned IDs are 5, 7, and 8 — these are not "zero-based identifiers + idCmdFirst" (that would yield 5, 6, 7 for three contiguous items). They're just arbitrary IDs that happen to start at idCmdFirst.

The original example is confusing because it uses 5 for both idCmdFirst and the first command ID, but the fix should clarify that, not introduce a formula that doesn't match the numbers. A clearer revision might be:

> For example, if *idCmdFirst* is set to 5 and you add three items with command identifiers 5, 7, and 8, the return value should be `MAKE_HRESULT(SEVERITY_SUCCESS, 0, 8 - 5 + 1)` — that is, the offset of the largest ID (8) minus *idCmdFirst* (5), plus one.

Since the author notes uncertainty, please have the Shell team (or Raymond Chen) verify before any wording is merged.

whindsaks:
> A clearer revision might be

This is not clearer, do NOT use the value 5 for both `idCmdFirst` and one of the the command IDs! Make `idCmdFirst` 3 or some other number that does not appear in whatever calculation is correct if `idCmdFirst` is not actually part of the calculation.

GrantMeStrength:
@copilot /review

GrantMeStrength (changes requested):
**Review: Request Changes**

The added parenthetical "(zero based identifiers + idCmdFirst)" is misleading. The command identifiers 5, 7, 8 in the example already include the `idCmdFirst` offset — they are the actual menu command IDs. The existing explanation with the formula `8 - 5 + 1 = 4` already correctly demonstrates how to compute the count.

The parenthetical implies these are pre-offset values, which would confuse readers.

whindsaks:
> which would confuse readers

You are already confusing me by using the same number for different things.

If it is like you are saying, then we should be able to change

 > For example, if <i>idCmdFirst</i> is set to 5 ...

to 

> For example, if <i>idCmdFirst</i> is set to 3 ... 

Correct? (This number 3 is then not used in the return value calculation at all, its only use would be in `InsertMenuItem` where `MENUITEMINFO::wID` would be `3(idCmdFirst) + 5(etc.)`)

GrantMeStrength:
@whindsaks You make a fair point — using the same number 5 for both `idCmdFirst` and a command count is confusing. I agree that changing the example to use a different value like 3 for `idCmdFirst` would make the distinction clearer. The PR author should revise the example so no number serves double duty. Requesting changes on the PR accordingly.

GrantMeStrength:
Closing due to inactivity — this change was sent back to the author but hasn't been updated in over two years. If it's still relevant, please feel free to reopen or submit a fresh PR against current content. Thanks for contributing!

whindsaks:
> this change was sent back to the author but hasn't been updated in over two years.

What do you mean? I'm the author and we talked about this just last week.

Please reopen this again. (And don't go around closing PRs as some form of cleanup!)

PR #1809, opened 2024-04-21
FindWindow does not set the last error

ChrisGuzak:
proven by this test

```cpp
    TEST_METHOD(DoesFindWindowSetLastError)
    {
        SetLastError(ERROR_TXF_METADATA_ALREADY_PRESENT);
        auto window = FindWindowW(L"Windows.UI.Core.CoreWindow", L"Start");
        auto err = GetLastError();
        cpp_unit::Assert::IsTrue(window != nullptr);
        cpp_unit::Assert::IsTrue(err == ERROR_TXF_METADATA_ALREADY_PRESENT);

        SetLastError(ERROR_TXF_METADATA_ALREADY_PRESENT);
        window = FindWindowW(L"Non existent", L"window");
        err = GetLastError();
        cpp_unit::Assert::IsTrue(window == nullptr);
        cpp_unit::Assert::IsTrue(err == ERROR_TXF_METADATA_ALREADY_PRESENT);
    }
```

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-findwindowa.md
+++ b/sdk-api-src/content/winuser/nf-winuser-findwindowa.md
@@ -91,7 +91,7 @@ Type: <b>HWND</b>
 
 If the function succeeds, the return value is a handle to the window that has the specified class name and window name.
 
-If the function fails, the return value is <b>NULL</b>. To get extended error information, call <a href="/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a>.
+This function does not modify the last error value.
 
 ## -remarks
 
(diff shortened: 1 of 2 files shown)


sdk-api-src/content/winuser/nf-winuser-findwindowa.md
```diff
@@ -91,7 +91,7 @@ Type: <b>HWND</b>
 
 If the function succeeds, the return value is a handle to the window that has the specified class name and window name.
 
-If the function fails, the return value is <b>NULL</b>. To get extended error information, call <a href="/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a>.
+This function does not modify the last error value.
```
jevansaks: I would keep the description about the return value like in findwindoww.
```suggestion
If the function fails, the return value is <b>NULL</b>. This function does not modify the last error value.
```

PR #1814, opened 2024-04-26
attempt at adding return type to getitemcount

antmor:
all of the ListView_* macros are missing a return type. I don't know if there is a simple way to automate this based on the header. But this at least does it for one function.


drewbatgit (approved):

drewbatgit:
Your changes have been verified against SDK headers and re-applied in PR #2186 (merged) with original author credit. Thank you for the contribution! Closing this PR as the conflict has been resolved.

PR #1815, opened 2024-04-28
Update StackFrame(_ex) pages

robotboy655:
Looks like the 2 pages were copypasted from the 64bit version without adjusting its contents correctly. This PR fixes that.

Affected pages:
* https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/ns-dbghelp-stackframe
* https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/ns-dbghelp-stackframe_ex

The c++ code example was directly coped from the windows header files.


robotboy655:

Karl-Bridge-Microsoft (approved):

PR #1817, opened 2024-04-29
Update ns-shellapi-shellexecuteinfow.md

ericlaw1979:
Correct documentation for NO_UI flag in the "W" version of this article after having fixed the "A" version this morning.

There probably should be a bot that recommends changing the "A" and "W" versions of the docs in the same PR.


jwmsft (approved):

jwmsft:
Thanks for the update. Bot suggestion has been passed along.

PR #1821, opened 2024-05-13
Update nf-wincrypt-cryptbinarytostringw.md

gtworek:
Added CRYPT_STRING_BASE64URI defined in wincrypt.h


alvinashcraft (approved):
This looks good. CryptBinaryToStringA should also be updated, I think. If you'd like to submit another PR for that, I will review it too. Otherwise, I'll add it to my to-do list.

PR #1823, opened 2024-05-18
Update nf-projectedfslib-prjstopvirtualizing.md

bundgaard:
The function type is void, which cannot return any other value than void.


bundgaard:
For your information this is the version I am running Projected Filesystem on.



From what I can see we have a non-exported method `int StopVirtualizationInstance(int)` is this the one the documentation is mentioning and we should use the `LoadLibrary` functionality instead to use ProjFS?
```
Edition	Windows 11 Pro
Version	23H2
Installed on	‎2023-‎04-‎24
OS build	22631.3593
Experience	Windows Feature Experience Pack 1000.22700.1003.0

```

PR #1824, opened 2024-05-22
Document use of `DXGI_OUTDUPL_FLAG` in `IDXGIOutput5::DuplicateOutput1()`

### LINKED ISSUE #128: Does `IDXGIOutput5::DuplicateOutput1()` take `enum DXGI_OUTDUPL_FLAG` in the `Flags` parameter?
We were investigating where `enum DXGI_OUTDUPL_FLAG` could be used (with its only known valid flag value `DXGI_OUTDUPL_COMPOSITED_UI_CAPTURE_ONLY = 1`), or if it was an accidental unused remnant.

@riverar [apparently reverse-engineered an example](https://github.com/microsoft/win32metadata/pull/1757#discussion_r1605614443) and [confirmed my initial hunch](https://github.com/microsoft/win32metadata/pull/1757#discussion_r1605484460) that this value is used for `Flags` in `IDXGIOutput5::DuplicateOutput1()`. The [documentation](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1) still suggests that the `Flags` value "must be zero" however.

Can anyone clear up whether this is true, and if so:

- Update the documentation;
- Maybe change the parameter type to directly take `DXGI_OUTDUPL_FLAG`, so that we don't have to hack this in `win32metadata` via `[AssociatedEnum]`?

After all `DXGI_OUTDUPL_FLAG` is an untyped enum which is assumed to be `INT` by default, mismatching the `UINT` ABI of the `Flags` parameter.

---

Note also that I'm trying to propagate this flag usage into `win32metadata` already via https://github.com/microsoft/win32metadata/pull/1911.

MarijnS95:
Fixes https://github.com/microsoft/DirectX-Headers/issues/128, cc @jenatali

This function is known to take values from the `DXGI_OUTDUPL_FLAG` enumeration.

Note that the type is still `UINT` (can we please change that without ABI repercussions), and there's no documentation page for `DXGI_OUTDUPL_FLAG` yet.

### DIFF
--- a/sdk-api-src/content/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1.md
+++ b/sdk-api-src/content/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1.md
@@ -62,9 +62,9 @@ A pointer to the Direct3D device interface that you can use to process the deskt
 
 ### -param Flags
 
-Type: <b>UINT</b>
+Type: <b>DXGI_OUTDUPL_FLAG</b>
 
-Reserved for future use; must be zero.
+A bitfield of <b>DXGI_OUTDUPL_FLAG</b> enumeration values describing the kind of capture surface to create.
 
 ### -param SupportedFormatsCount [in]
 
--- a/sdk-api-src/content/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1.md
+++ b/sdk-api-src/content/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1.md
@@ -62,7 +62,7 @@ A pointer to the Direct3D device interface that you can use to process the deskt
 
 ### -param Flags
 
-Type: <b>DXGI_OUTDUPL_FLAG</b>
+Type: <b>UINT</b>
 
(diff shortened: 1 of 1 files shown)


sdk-api-src/content/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1.md
```diff
@@ -62,9 +62,9 @@ A pointer to the Direct3D device interface that you can use to process the deskt
 
 ### -param Flags
 
-Type: <b>UINT</b>
+Type: <b>DXGI_OUTDUPL_FLAG</b>
```
jenatali: I don't think this `Type:` field can change, but the description looks good to me.

MarijnS95: Description was randomly thought up, unsure if you want to go with the "surface" wording.

jenatali (approved):
Seems fine to me, I'll let the doc writer adjust wording as necessary.

stevewhims (approved):

PR #1825, opened 2024-05-22
Update nf-ncrypt-ncryptcreatepersistedkey.md

emilyzhang47:
Adding documentation for NCRYPT_USE_PER_BOOT_KEY_FLAG


alvinashcraft (approved):
The change looks good here. Let me know if the new flag needs to also be added to the table on [NCryptImportKey](https://learn.microsoft.com/windows/win32/api/ncrypt/nf-ncrypt-ncryptimportkey).

PR #1827, opened 2024-05-24
ne-objbase-coinit.md: fix COINIT_MULTITHREADED

dlech:
COINIT_MULTITHREADED was listed after COINIT_APARTMENTTHREADED without an initializer which according to the C language would make it have a value of 0x3 which is incorrect. The actually value is 0x0 (via the COINITBASE_MULTITHREADED enum value)


stevewhims (approved):
It's good to have that constant in there, but we can't rearrange the order in that way (they'd still render in alpha order).

PR #1829, opened 2024-06-06
Update nf-d3d11-id3d11videocontext-decoderbeginframe.md

HuaChunbo:
Remove the redundant function name to avoid confusion.


stevewhims:
Yes, I think removing this is ok (the two titles are still unique). The title for ID3D11VideoContext3::DecoderBeginFrame1 should say something like "Starts a decoding operation to decode a video frame, with histogram options". That's how we should be disambiguating titles, not just sticking the API's name at the end of the description. Which is, as you say, confusing.

stevewhims (approved):

PR #1841, opened 2024-06-17
Update nf-winerror-hresult_facility.md

andrewimcclement:
Bit shift is in wrong direction.


andrewimcclement:
From my local C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared\winerror.h:
image

stevewhims (approved):

PR #1842, opened 2024-06-19
Update nf-winbase-writeprivateprofilestringw.md

DJm00n:
Fix wording according to the @oldnewthing blog post [Can INI files be Unicode? Yes, they can, but it has to be your idea](https://devblogs.microsoft.com/oldnewthing/20240606-00/?p=109861)


GrantMeStrength:
Thanks!

PR #1844, opened 2024-06-24
Updated table to include SHA1_SHA256 & SHA256 hashes

Xelu86:


alvinashcraft (approved):
This looks good. Matches up with the values in the current header. Thanks!

mmwuyan (commented):
This looks good to me.

PR #1854, opened 2024-07-09
Update nf-winerror-hresult_severity.md

colejohnson66:
Fix HTML escaping bug


colejohnson66:
Other pages also have the same issue, such as [`HRESULT_SEVERITY`](https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-hresult_severity), [`SCODE_CODE`](https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-scode_code), etc.

stevewhims (approved):

PR #1858, opened 2024-07-12
Fix CreateWindowW() Return value section

chrisn:
This page currently has both a **Return value** and a **Returns** section.


jwmsft (approved):

jwmsft:
Thanks!

PR #1859, opened 2024-07-17
Update nf-aclapi-getnamedsecurityinfow.md

xavetar:
Fix error in documentation

### DIFF
--- a/sdk-api-src/content/aclapi/nf-aclapi-getnamedsecurityinfow.md
+++ b/sdk-api-src/content/aclapi/nf-aclapi-getnamedsecurityinfow.md
@@ -103,7 +103,7 @@ This parameter is required if any one of the <i>ppsidOwner</i>, <i>ppsidGroup</i
 
 If the function succeeds, the return value is ERROR_SUCCESS.
 
-If the function fails, the return value is a nonzero error code defined in WinError.h.
+If the function succeeds, the return value is a nonzero error code defined in WinError.h (87).
 
 ## -remarks
 


alvinashcraft (changes requested):
This change is not correct for all cases. There are failure conditions that return a nonzero error code. Looking at the code, ERROR_INVALID_PARAMETER is just one of the possible return codes that would not indicate the function succeeded.

xavetar:
> This change is not correct for all cases. There are failure conditions that return a nonzero error code. Looking at the code, ERROR_INVALID_PARAMETER is just one of the possible return codes that would not indicate the function succeeded.

Maybe you can check it first? Do I need to send a test code? When the function (GetNamedSecurityInfoW/A) is executed successfully, it returns:

1) Code 0 (ERROR_SUCCESS)
2) GetLastError() -> 87 (DWORD).

And I don't quite understand where the error is possible if the result of the function itself is positive, and the error code is negative. This is not my logic, but Microsoft's logic, I had to spend a lot of time to understand that the error is not where I am looking for it and that the description of the function and its operation do not correspond to the documentation.

I checked the error code by the result of GetLastError(), which always sets the code equal to 87 when ERROR_SUCCESS is successfully executed. Perhaps the error occurs inside advapi32, in any case, these sections of the documentation require modification and correction.

alvinashcraft:
> And I don't quite understand where the error is possible if the result of the function itself is positive, and the error code is negative. This is not my logic, but Microsoft's logic, I had to spend a lot of time to understand that the error is not where I am looking for it and that the description of the function and its operation do not correspond to the documentation.
> 

But the paragraph you've changed is the case where the function results is NOT positive. We don't want to remove this part of the docs because the function result may not be successful. If you think some more docs need to be added for the success case, that's fine, but don't remove the failure case.

xavetar:
> > And I don't quite understand where the error is possible if the result of the function itself is positive, and the error code is negative. This is not my logic, but Microsoft's logic, I had to spend a lot of time to understand that the error is not where I am looking for it and that the description of the function and its operation do not correspond to the documentation.
> 
> But the paragraph you've changed is the case where the function results is NOT positive. We don't want to remove this part of the docs because the function result may not be successful. If you think some more docs need to be added for the success case, that's fine, but don't remove the failure case.

Technically, should I add that if GetNamedSecurityInfoW/A fails, it returns something other than ERROR_SUCCESS?

> If the function fails, the return value is a nonzero.

Then, theoretically, should I also add:

1) That the error code of GetLastError() should be ignored
2) Not mentioned at all?
3) Can the error code be anything?
4) Error code of 87 or 0 acceptable?
5) Error code of 87 or 0 be considered a success function call?
6) Error code other than 87 or 0 can be considered an failed function call?

alvinashcraft:
The docs as they are written today read correctly to me.

If the functions are successful, they return ERROR_SUCCESS, which is code 0. If they fail, they return some other code, which can be referenced in winerror.h. 

They don't return 87, do they? You said that's the value you got after calling GetLastError() when the function was successful. The docs don't typically recommend calling GetLastError when a function returns a success code. It's something that's called when there is a failure.

What is currently documented is what is usually documented for return values. Tracking down every possible error code that could be returned by functions in the Win32 headers would be fantastic, but maintaining the correctness of those docs over time without automation would be nearly impossible.

xavetar:
> They don't return 87, do they? You said that's the value you got after calling GetLastError() when the function was successful. The docs don't typically recommend calling GetLastError when a function returns a success code. It's something that's called when there is a failure.

Um, if successful, the function can return an error code via GetLastError() even when there is no error? Okay, then why mention the Winerror.h at all?

How can a negative error code indicating that an error occurred indicate successful execution?

If this happens inside other WinAPI functions, then what is the practical point of this API if it can return an error when there is none?

This break logic and distort perception. Like 1+1 in regular mathematics can be equal to 0 or 1? Is this some kind of quantum mechanics or non-ordinary logic?

xavetar:
If no one is going to fix the problem and clarify it in the documentation, I delete the repository, the problem is closed.

PR #1860, opened 2024-07-17
Update nf-aclapi-getnamedsecurityinfoa.md

xavetar:
Fix error in documentation

### DIFF
--- a/sdk-api-src/content/aclapi/nf-aclapi-getnamedsecurityinfoa.md
+++ b/sdk-api-src/content/aclapi/nf-aclapi-getnamedsecurityinfoa.md
@@ -103,7 +103,7 @@ This parameter is required if any one of the <i>ppsidOwner</i>, <i>ppsidGroup</i
 
 If the function succeeds, the return value is ERROR_SUCCESS.
 
-If the function fails, the return value is a nonzero error code defined in WinError.h.
+If the function succeeds, the return value is a nonzero error code defined in WinError.h (87).
 
 ## -remarks
 


alvinashcraft (changes requested):
Same issue as the other PR for GetNamedSecurityInfoW.
This change is not correct for all cases. There are failure conditions that return a nonzero error code. Looking at the code, ERROR_INVALID_PARAMETER is just one of the possible return codes that would not indicate the function succeeded.

xavetar:
> Same issue as the other PR for GetNamedSecurityInfoW. This change is not correct for all cases. There are failure conditions that return a nonzero error code. Looking at the code, ERROR_INVALID_PARAMETER is just one of the possible return codes that would not indicate the function succeeded.

Maybe you can check it first? Do I need to send a test code? When the function (GetNamedSecurityInfoW/A) is executed successfully, it returns:

1) Code 0 (ERROR_SUCCESS)
2) GetLastError() -> 87 (DWORD).

And I don't quite understand where the error is possible if the result of the function itself is positive, and the error code is negative. This is not my logic, but Microsoft's logic, I had to spend a lot of time to understand that the error is not where I am looking for it and that the description of the function and its operation do not correspond to the documentation.

I checked the error code by the result of GetLastError(), which always sets the code equal to 87 when ERROR_SUCCESS is successfully executed. Perhaps the error occurs inside advapi32, in any case, these sections of the documentation require modification and correction.

xavetar:
If no one is going to fix the problem and clarify it in the documentation, I delete the repository, the problem is closed.

PR #1864, opened 2024-07-30
IPreviewHandler: add interface to Inizialize

davidegiacometti:
The statement
```
Additionally, this method should be called only after Initialize, Initialize, or Initialize has been called.`
```
doesn't allow to differentiate the 3 Initialize at a glance.
Added the interface to the 3 Inizialize.
```
Additionally, this method should be called only after IInitializeWithItem::Initialize, IInitializeWithStream::Initialize, or IInitializeWithFile::Initialize has been called.`
```


jwmsft (approved):

jwmsft:
@davidegiacometti, thanks for this update.

jwmsft (approved):

PR #1868, opened 2024-08-07
Document RmGetList returning ERROR_ACCESS_DENIED

yaakov-h:
If one calls `RmRegisterResources` with a directory path, `RmGetList` will fail subsequently with `ERROR_ACCESS_DENIED`.

I first found this 8 years ago and based on some use of RM I had this week, it seems that Windows has not changed.

https://blog.yaakov.online/failed-experiment-what-processes-have-a-lock-on-this-folder/


yaakov-h:
@microsoft-github-policy-service agree company="WiseTech Global"

drewbatgit:
Thanks for contributing this info to the documentation.

PR #1879, opened 2024-08-21
Correct parameter description

Whitecx:
A parameter in GetIpNetTable2 docs is described as a MIB_IPNET_TABLE2 structure, but it is a pointer to one.


stevewhims (approved):

stevewhims:
Let's just say pointer to a pointer.

PR #1880, opened 2024-08-22
Update nf-bcrypt-bcryptenumalgorithms.md

ccalik:
Add key derivation operation flag


alvinashcraft (approved):
Thanks for adding this. The value matches what I'm seeing in the header, so I'll get this merged.

PR #1882, opened 2024-08-30
Document return type for GetNextWindow()

Pavel-Cherentsov:
Put return type **_HWND_** instead of **_void_**.

Current **_void_** and _Return value **None**_ look incorrect:
image


jwmsft:
Thanks for submitting this update. This is a duplicate of PR #1877, so I am closing this one.

PR #1886, opened 2024-09-02
Update ns-winbase-copyfile2_extended_parameters_v2.md

RatinCN:
Fix incorrect table style.

Before:
image

Proposal change be like:
image


GrantMeStrength:
Thanks for the fix!

PR #1887, opened 2024-09-04
Update nf-fileapi-gettemppath2a.md

hihayak:
"Windows 10 build 20348" doesn't exist and it looks confusing.

### DIFF
--- a/sdk-api-src/content/fileapi/nf-fileapi-gettemppath2a.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-gettemppath2a.md
@@ -20,7 +20,7 @@ req.lib: Kernel32.lib
 req.max-support: 
 req.namespace: 
 req.redist: 
-req.target-min-winverclnt: Windows 11 Build 22000
+req.target-min-winverclnt: Windows 10 Build 19044
 req.target-min-winversvr: Windows Server Build 20348
 req.target-type: 
 req.type-library: 
--- a/sdk-api-src/content/fileapi/nf-fileapi-gettemppath2a.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-gettemppath2a.md
@@ -20,7 +20,7 @@ req.lib: Kernel32.lib
 req.max-support: 
 req.namespace: 
 req.redist: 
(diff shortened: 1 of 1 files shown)


ChrisDenton:
I believe 20348 is Windows Server 2022.

alvinashcraft:
> I believe 20348 is Windows Server 2022.

That's correct. Somebody plugged the server build number in the client field. That release was Aug 2021, which would equate to Windows 10 21H2 or build 19044.

sdk-api-src/content/fileapi/nf-fileapi-gettemppath2a.md
```diff
@@ -20,7 +20,7 @@ req.lib: Kernel32.lib
 req.max-support: 
 req.namespace: 
 req.redist: 
-req.target-min-winverclnt: Windows 10 Build 20348
+req.target-min-winverclnt: Windows 11 Build 22000
```
alvinashcraft: I think this would be the correct minimum client version:

```suggestion
req.target-min-winverclnt: Windows 10 Build 19044
```

alvinashcraft (approved):

hihayak:
@alvinashcraft, thank you for taking a look at this. It seems you changed the fix as below.
"Windows 10 Build 19044"

I think this is not correct. This function is available in the build 20348 or later. Thus, Windows 10 doesn't support this function. You can confirm it by checking the result of "dumpbin /exports c:\Windows\System32\kernel32.dll | findstr /I gettemppath".

Could you please consider reverting this to my original amendment? Also, could you update GetTempPath2W in the same manner?

alvinashcraft:
Thanks, @hihayak. I'm checking with the dev team for fileapi.h before I make any other changes. When they confirm the correct client version number, I'll create a new PR to fix GetTempPath2A and GetTempPath2W. I'll circle back with a comment here when the PR is complete.

alvinashcraft:
Hi @hihayak. I finally got confirmation from the product team. This change was initially made on the server OS. It was later included in the client, but it's only in Windows 11 21H2 and later. That's Build 22000 or later for the client. I just merged a PR to reflect this version info, and it should be live soon. Thanks for your patience!

PR #1888, opened 2024-09-04
Specify which offset _interlockedbittestandset uses

anrossi:
Update the documentation to explain the offset is from the least-significant bit, and that it's zero-indexed, so 0 is the least-significant bit, and 31 is the most significant bit.
The documentation for related APIs should also probably be updated, but I wanted to get feedback on this change first.


GrantMeStrength:
Thanks for the update.

PR #1900, opened 2024-09-20
Update ns-winnt-image_debug_directory.md

DJm00n:
Link to the full table of possible `IMAGE_DEBUG_TYPE_*` values.


GrantMeStrength:
Thanks, good way to avoid duplication.

PR #1915, opened 2024-10-16
Escape underscores in nf-lzexpand-lzopenfilew.md

zooba:
Without escapes, the underscores are invisible and the text is shown with italics. The other underscores are fine because they are inside HTML elements and so are treated differently by the Markdown engine.


alvinashcraft (approved):
Thanks for fixing this. The mixed markdown and HTML leftover from the MSDN conversion has created some inconsistencies like this. Looks good.

PR #1919, opened 2024-10-21
Publish the WLDP APIs: GetApplicationSetting

jgeurten:
Adding three new public APIs in the WLDP library: 

1. WldpGetApplicationSettingBoolean
2. WldpGetApplicationSettingStringList
3. WldpGetApplicationSettingStringSet


bowerscd-corp (changes requested):
Will review more in depth in a bit, just preventing merge without approval.

PR #1925, opened 2024-10-29
Update nf-fileapi-getvolumepathnamesforvolumenamew.md

gtworek:
TCHAR replaced with WCHAR, as only WCHAR is correct for XxxxxW function.


alvinashcraft (approved):
This is a good catch. Thank you. I'm going to make a note to look for this issue in other "W" versions of fn's.

PR #1931, opened 2024-11-13
Update nf-shlwapi-strformatbytesizeex.md

iPhilip:
Removed confusing sentence:

> In Windows 10, size is reported in base 10 rather than base 2. For example, 1 KB is 1000 bytes rather than 1024.

The statement is only true for marketing materials ([Link](https://answers.microsoft.com/en-us/windows/forum/all/windows-10-file-explorer-unit-confusion/fed69f40-c448-4028-a011-cb0780215a63)).

### DIFF
--- a/sdk-api-src/content/shlwapi/nf-shlwapi-strformatbytesizeex.md
+++ b/sdk-api-src/content/shlwapi/nf-shlwapi-strformatbytesizeex.md
@@ -107,9 +107,7 @@ The following table illustrates how this function converts a numeric value into 
 <td>1.99 GB</td>
 </tr>
 </table>
- 
 
-In Windows 10, size is reported in base 10 rather than  base 2. For example, 1 KB is 1000 bytes rather than 1024.
 
 ## -see-also
 


riverar (commented):
I can't reproduce the documented behavior on Windows 10, Windows 11, or Windows vNext. Might be worth checking Windows XP and Windows 7.

iPhilip:
> I can't reproduce the documented behavior on Windows 10, Windows 11, or Windows vNext. Might be worth checking Windows XP and Windows 7.

```
#include <windows.h>
#include <shlwapi.h>
#include <stdio.h>

#pragma comment(lib, "shlwapi.lib")

int main(void) {
    WCHAR buffer[64];
    ULONGLONG bytes = 1024;

    // Format 1000 bytes with default settings
    StrFormatByteSizeEx(
        bytes,
        SFBS_FLAGS_ROUND_TO_NEAREST_DISPLAYED_DIGIT,
        buffer,
        64
    );

   wprintf(L"1024 bytes = %s\n", buffer);  // 1024 bytes = 1.00 KB

    return 0;
}
```

The above code clearly demonstrates that 1KB is 1024 bytes - not 1000 bytes.

The wording in the current documentation page:

> For example, 1 KB is 1000 bytes rather than 1024.

contradicts the above.

GrantMeStrength (changes requested):
Before removing this sentence, the Windows shell team should confirm whether `StrFormatByteSizeEx` actually changed its base in Windows 10.

The PR body links to a user forum discussion rather than authoritative documentation. File Explorer's display of file sizes in base-10 in Windows 10 is real OS behavior, and `StrFormatByteSizeEx` may follow the same convention — removing this sentence without confirmation risks silently deleting accurate information.

Please get a sign-off from someone on the Shell/shlwapi team who can confirm the actual runtime behavior of this function on Windows 10+. If the sentence is wrong, it would also be worth replacing it with an accurate description rather than removing it entirely.

GrantMeStrength:
@copilot /review

GrantMeStrength (commented):
**Review: Needs Team Verification**

This PR removes the statement: *"In Windows 10, size is reported in base 10 rather than base 2."*

If this statement was accurate, removing it would leave developers unaware of a platform behavior difference. If it was inaccurate, it should be removed.

Could the team verify whether StrFormatByteSizeEx actually changed its base-10/base-2 behavior in Windows 10? This requires checking against actual API behavior before merging.

GrantMeStrength:
@riverar Thanks for testing this. Since you can't reproduce the documented behavior on any modern Windows version, this PR's claim seems unverifiable. I'm inclined to close this unless the author can provide specific repro steps or a Windows version where this occurs. The existing documentation may already be correct.

GrantMeStrength:
Closing this PR — the change has already been applied independently. The confusing "base 10" sentence referenced in this PR has been removed from the current version of the documentation. Thank you for the contribution!

PR #1936, opened 2024-11-18
Update nf-uxtheme-setwindowtheme.md

RatinCN:
An empty string should not contain such a space

### DIFF
--- a/sdk-api-src/content/uxtheme/nf-uxtheme-setwindowtheme.md
+++ b/sdk-api-src/content/uxtheme/nf-uxtheme-setwindowtheme.md
@@ -86,7 +86,7 @@ If this function succeeds, it returns <b>S_OK</b>. Otherwise, it returns an <b>H
 The theme manager retains the <i>pszSubAppName</i> and the <i>pszSubIdList</i> associations through the lifetime of the window, even if visual styles subsequently change. The window is sent a <a href="/windows/desktop/winmsg/wm-themechanged">WM_THEMECHANGED</a> message at the end of a <b>SetWindowTheme</b> call, so that the new visual style can be found and applied.
 
 
-When <i>pszSubAppName</i> and <i>pszSubIdList</i> are <b>NULL</b>, the theme manager removes the previously applied associations. You can prevent visual styles from being applied to a specified window by specifying an empty string, (L" "), which does not match any section entries. 
+When <i>pszSubAppName</i> and <i>pszSubIdList</i> are <b>NULL</b>, the theme manager removes the previously applied associations. You can prevent visual styles from being applied to a specified window by specifying an empty string, (L""), which does not match any section entries. 
 
 
 #### Examples


sdk-api-src/content/uxtheme/nf-uxtheme-setwindowtheme.md
```diff
@@ -86,7 +86,7 @@ If this function succeeds, it returns <b>S_OK</b>. Otherwise, it returns an <b>H
 The theme manager retains the <i>pszSubAppName</i> and the <i>pszSubIdList</i> associations through the lifetime of the window, even if visual styles subsequently change. The window is sent a <a href="/windows/desktop/winmsg/wm-themechanged">WM_THEMECHANGED</a> message at the end of a <b>SetWindowTheme</b> call, so that the new visual style can be found and applied.
 
 
-When <i>pszSubAppName</i> and <i>pszSubIdList</i> are <b>NULL</b>, the theme manager removes the previously applied associations. You can prevent visual styles from being applied to a specified window by specifying an empty string, (L" "), which does not match any section entries. 
+When <i>pszSubAppName</i> and <i>pszSubIdList</i> are <b>NULL</b>, the theme manager removes the previously applied associations. You can prevent visual styles from being applied to a specified window by specifying an empty string, (L""), which does not match any section entries.
```
riverar, line 89: I think this space is deliberate and we should correct the text instead. After a quick glance, it does appear `uxtheme!SetWindowTheme` behaves differently if `""`, `" "`, or `nullptr` is passed in.

```suggestion
When <i>pszSubAppName</i> and <i>pszSubIdList</i> are <b>NULL</b>, the theme manager removes the previously applied associations. You can prevent visual styles from being applied to a specified window by specifying a string with a space, (L" "), which does not match any section entries. 
```

riverar (changes requested):

sdk-api-src/content/uxtheme/nf-uxtheme-setwindowtheme.md
RatinCN: If pass `L""`, `SetWindowTheme` will assign `L"$"` to the property, which does not match anything.
And `L" "` does not match anything too, because no theme atom named " ".
`SetWindowTheme` compares income string with `L""` explicitly, and uses `L"$"` replace it. So I think this should be the preset input for this function, rather than `L" "`.

riverar: The current implementation has the `$` logic, but I wonder if that's supported downlevel with older releases. If so, then I agree!

RatinCN: As far as I know, Windows keeps this behavior since NT5. Imoh, the key is it compares `L""`, so this should be the preset input, whatever it replaces empty string by `L"$"` or `L" "` or others. Thanks for your attention : )

riverar: Sounds good to me, feel free to mark as resolved.

riverar (approved):

jwmsft (approved):

PR #1941, opened 2024-12-04
Update nf-slpublic-slgetwindowsinformationdword.md to document SL_E_VALUE_NOT_FOUND

dmachaj:
This is a useful return code to call out specifically as useful to callers of this API.


michbern-ms (approved):
Thanks, David, for adding this error code.  I confirmed the value of SL_E_VALUE_NOT_FOUND in the OS source code.

PR #1949, opened 2024-12-16
IMAGE_FILE_HEADER - update machine info with link to full list

wesinator:
update `machine` field info with a full list of the machine types
link is to the doc found at https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/Debug/pe-format.md

[NO_TRAIN]::


GrantMeStrength:
Thanks!

PR #1951, opened 2025-01-03
Update nf-winuser-getmenuitemcount.md

iPhilip:
It's not clear why the `hMenu` parameter is optional.


jwmsft:
@iPhilip, thanks for your PR submissions.
Parameters are auto generated from the header file. WinUser.h defines hMenu as optional (although I'm not sure why it does). 
From WinUser.h:
```
int
WINAPI
GetMenuItemCount(
    _In_opt_ HMENU hMenu);
```    
The current doc seems to be correct, so I will close this PR.

PR #1953, opened 2025-01-07
RawInput: Buffer pointer should be aligned on a DWORD (32-bit) boundary

DJm00n:
Buffer pointer should be aligned on a **DWORD** (32-bit) boundary.


riverar (changes requested):
Microsoft is saying that the starting address of each `RAWINPUT` structure within the buffer must be aligned to a pointer boundary. The pointer itself doesn’t need special alignment.

Memory allocated with `malloc` or `new` will already handle this, so it's not clear why this warning is here. My guess is someone got burned after using a custom allocator (e.g., on say, the Xbox) and added a note for future developers.

Karl-Bridge-Microsoft (approved):

PR #1971, opened 2025-02-05
Update CodecAPI to limit the scope of legacy usage to Dshow and add new URLs to DX12 Encode

RohitAthavale:
This PR makes the following changes

1. Clarify that Codecapi usage with DirectShow is a legacy feature.
2. Remove links to Direct Show Encoder pages, instead point to more active DX12 Encode Pages

### DIFF
--- a/sdk-api-src/content/strmif/nn-strmif-icodecapi.md
+++ b/sdk-api-src/content/strmif/nn-strmif-icodecapi.md
@@ -51,7 +51,7 @@ api_name:
 
 ## -description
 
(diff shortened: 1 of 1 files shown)


stevewhims (approved):

stevewhims:
Note that the instance of this Note that we've updated here is just 1 of 4,873 identical Notes across the documentation. The other 4,872 are as yet unchanged.

RohitAthavale:
@stevewhims Could you tell me how you queried or grep'ed for these 4873 identical Notes. If this change snippet looks fine, I'll try to write a script to bulk edit these.

drewbatgit:
@RohitAthavale Thanks for your contribution to the docs. After further review your changes, we will keep your change to the See Also section, but we are going to roll back the change to the note, so there is no need to make any additional updates.

PR #1974, opened 2025-02-14
Clarify DML_ROUNDING_MODE

fdwr:
The behavior was unclear (arose in WebNN discussion about rounding). So:
- clarify with examples verified via [DxDispatch](https://github.com/microsoft/DirectML/blob/master/DxDispatch/doc/Guide.md).
- fix descriptions, as *toward-zero* does not always round downward (that is *floor*) nor does *halves-to-toward-infinity* always round upward (that is *ceil*).

> [!NOTE]
> In hindsight, I named `DML_ROUNDING_MODE_TOWARD_INFINITY` poorly and should have called it `DML_ROUNDING_MODE_HALVES_TOWARD_INFINITY`, as it isn't the opposite of `DML_ROUNDING_MODE_TOWARD_ZERO` 🤦‍♂️.

### DIFF
--- a/sdk-api-src/content/directml/ne-directml-dml_rounding_mode.md
+++ b/sdk-api-src/content/directml/ne-directml-dml_rounding_mode.md
@@ -52,16 +52,22 @@ Defines constants that specify a rounding mode.
 
 ### -field DML_ROUNDING_MODE_HALVES_TO_NEAREST_EVEN
 
-Specifies rounding halves to the nearest even integer.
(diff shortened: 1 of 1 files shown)


fdwr:
Oh, I see the [operator desc page](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_element_wise_round_operator_desc) also describes the mode enums and has more clarity (whereas I naturally searched for the enum name to learn about the enum 😉). Well the example table is still useful either way.

stevewhims (approved):

PR #1996, opened 2025-03-21
Fixed typo and documented NTE_BAD_HASH return value

icnocop:


alvinashcraft (approved):
Looks good. Thanks for fixing the typo and adding the additional info about return value.

PR #2000, opened 2025-04-09
HOOKPROC never mentions CWPSTRUCT

edcdecl:
HOOKPROC can receive both CWPSTRUCT and CWPRETSTRUCT in LPARAM, but CWPSTRUCT is not once mentioned within the documentation.


drewbatgit (approved):

PR #2003, opened 2025-04-23
Update nn-tsvirtualchannels-iwtsplugin.md

snakex2000:
Trying to resolve the broken link at bottom of page for VirtualChannelGetInstance function.

Docs should point to same place as the earlier reference in these docs.


drewbatgit (approved):

PR #2007, opened 2025-04-24
Update nf-mfmediaengine-imfextendeddrmtypesupport-istypesupportedex.md

cjrog:
Update docs after support has been added for the -clearlead suffix starting in Windows 10 OS Build 19045.5796 and later.

This support has been backported from Windows 11 to support clearlead with HWDRM on Windows 10.


riverar:
Says who?

sdk-api-src/content/mfmediaengine/nf-mfmediaengine-imfextendeddrmtypesupport-istypesupportedex.md
```diff
@@ -117,7 +117,7 @@ The following table lists the supported individual feature queries, organized by
...
+|6a     |Decryption|encryption-type         |“cenc” or “cbcs”, with optional "-clearlead" suffix. |Is this encryption type supported for decryption with the specified codec / key-system? If value is unspecified, default value of "cenc" is used. Starting with Windows 11 Build 22621 and Windows 10 Build 19045.5796 or later, the suffix "-clearlead" is supported. When "-clearlead" is appended to the encryption type value, support for using clear content at the very beginning of encrypted content is also requested. Clear content at the beginning of the content speeds up the time to presenting the first frame. If "-clearlead" is added to the encryption-type, the version number of the requested codec will be checked. The AV1 and VP9 codecs will be checked for major version 2, and HEVC will be checked for v2.0.53421 or greater.       |N       |
```
riverar, line 120: If this is verified to be true, we should probably flip the versioning to read more naturally here:

```suggestion
|6a     |Decryption|encryption-type         |“cenc” or “cbcs”, with optional "-clearlead" suffix. |Is this encryption type supported for decryption with the specified codec / key-system? If value is unspecified, default value of "cenc" is used. The "-clearlead" suffix is supported in Windows 10 (build 19045.5796 or later) and in Windows 11 (build 22621 or later). When "-clearlead" is appended to the encryption type value, support for using clear content at the very beginning of encrypted content is also requested. Clear content at the beginning of the content speeds up the time to presenting the first frame. If "-clearlead" is added to the encryption-type, the version number of the requested codec will be checked. The AV1 and VP9 codecs will be checked for major version 2, and HEVC will be checked for v2.0.53421 or greater.       |N       |
```

riverar:
@drewbatgit What about the edits suggested above?

drewbatgit:
@riverar The existing wording matches our conventions in the documentation, so your suggested change was not adopted. "Reading naturally" is a subjective assessment. So, while we can't please everyone, we try to use consistent phrasing for things like this.

riverar:
I find it very hard to believe that a backwards range of Windows versions matches any conventions.

PR #2008, opened 2025-04-30
Update misleading informations in NtNotifyChangeMultipleKeys

xmine64:
During my analysis, I put so much time wondering why my call to NtNotifyChangeMultipleKeys returns with `STATUS_INVALID_PARAMETER` and `STATUS_INVALID_OBJECT_NAME` while everything is correct, according to the documentations, just to find out there are misleading information in it. I'm making this PR so other people won't be confused like me.

### DIFF
--- a/sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
+++ b/sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
@@ -62,11 +62,11 @@ A handle to an open key. The handle must be opened with the <b>KEY_NOTIFY</b> ac
 
 ### -param Count [in, optional]
 
-The number of subkeys under the key specified by the <i>MasterKeyHandle</i> parameter. This parameter must be 1.
+The number of keys objects provided in the <i>SubordinateObjects</i> parameter. This parameter must be 1.
 
 ### -param SubordinateObjects [in, optional]
 
-Pointer to an array of <a href="/windows-hardware/drivers/ddi/content/wudfwdm/ns-wudfwdm-_object_attributes">OBJECT_ATTRIBUTES</a> structures, one for each subkey.   This array can contain one <b>OBJECT_ATTRIBUTES</b> structure.
+Pointer to an array of <a href="/windows-hardware/drivers/ddi/content/wudfwdm/ns-wudfwdm-_object_attributes">OBJECT_ATTRIBUTES</a> structures, one for each key.   This array can contain one <b>OBJECT_ATTRIBUTES</b> structure and you can't provide a key in the same hive as the <i>MasterKeyHandle</i> key.
 
 ### -param Event [in, optional]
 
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):

sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
```diff
@@ -62,23 +62,23 @@ A handle to an open key. The handle must be opened with the <b>KEY_NOTIFY</b> ac
 
 ### -param Count [in, optional]
 
-The number of subkeys under the key specified by the <i>MasterKeyHandle</i> parameter. This parameter must be 1.
+The number of keys objects provided in the <i>SubordinateObjects</i> parameter. This parameter must be 1.
 
 ### -param SubordinateObjects [in, optional]
 
-Pointer to an array of <a href="/windows-hardware/drivers/ddi/content/wudfwdm/ns-wudfwdm-_object_attributes">OBJECT_ATTRIBUTES</a> structures, one for each subkey.   This array can contain one <b>OBJECT_ATTRIBUTES</b> structure.
+Pointer to an array of <a href="/windows-hardware/drivers/ddi/content/wudfwdm/ns-wudfwdm-_object_attributes">OBJECT_ATTRIBUTES</a> structures, one for each key.   This array can contain one <b>OBJECT_ATTRIBUTES</b> structure and must not be a key in the same hive as the <i>MasterKeyHandle</i> key.
```
HBelusca, line 69: If these aren't in the same hive as the master root key, how then could they be "subkeys" of a given registry key, as the summary description of this function claims they are?
What then would be the usefulness of the function altogether if this weren't the case?

sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
```diff
@@ -62,23 +62,23 @@ A handle to an open key. The handle must be opened with the <b>KEY_NOTIFY</b> ac
...
 
 ### -param Event [in, optional]
 
 A handle to an event created by the caller. If <i>Event</i> is not <b>NULL</b>, the caller waits until the operation succeeds, at which time the event is signaled.
 
 ### -param ApcRoutine [in, optional]
 
-A pointer to an asynchronous procedure call (APC) function supplied by the caller. If <i>ApcRoutine</i> is not <b>NULL</b>, the specified APC function executes after the operation completes.
+A pointer to an asynchronous procedure call (APC) function supplied by the caller. If <i>ApcRoutine</i> is not <b>NULL</b>, the specified APC function executes after the operation completes. A <a href="/windows-hardware/drivers/ddi/wdm/ns-wdm-_work_queue_item">WORK_QUEUE_ITEM</a> must be provided instead of ApcRoutine in the <i>ZwNotifyChangeMultipleKeys</i> variant.
```
HBelusca, line 77: The proposed change is strange. Normally zw routines are == NT ones except for the detour back into the syscall table that changes the PreviousMode to KerneMode.
(When being invoked from user-mode via ntdll, both are the same thing.)
Are you suggesting that the parameter purpose completely changes altogether?

sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
xmine64: Being a subkey isn't a requirement, I tested the API on \Registry\User\.Default\TestKey while the master key was \Registry\Machine\Software\TestKey and it worked. While being a subkey is not a requirement, it's possible to load a hive in a non-root key so that way a key would have a subkey from another hive. As far as I know that's how BCD and SAM database are loaded in to the registry, they're not a root key like HKLM but a subkey of a parent key, but they're from a separate hive compared to their parent.

sdk-api-src/content/winternl/nf-winternl-ntnotifychangemultiplekeys.md
xmine64: The said parameters are just pointers, and they can be treated differently. This behavior was documented in ZwNotifyChangeKey but not here.

PR #2012, opened 2025-05-02
user32: Remove incorrect remark from TileWindows and CascadeWindows

nir9:
Problem: The documentation of TileWindows and CascadeWindows contains a remark claiming that after calling these functions, the maximized windows are restored to their original size. This is a partially correct claim; while the windows are indeed restored, they are not restored to the original size, but rather to a specific size appropriate for the tiling/cascade pattern.

Solution: Remove the incorrect remark from both doc pages.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-cascadewindows.md
+++ b/sdk-api-src/content/winuser/nf-winuser-cascadewindows.md
@@ -124,9 +124,7 @@ If the function fails, the return value is zero. To get extended error informati
 
 ## -remarks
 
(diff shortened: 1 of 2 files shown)


GrantMeStrength (approved):
Correct removal. The claim that these functions restore maximized windows 'to their previous size' is inaccurate — the functions restore them to their default restored size, which is not necessarily the size they had before being maximized. Also fixes missing EOF newlines in both files.

GrantMeStrength:
@copilot /review

GrantMeStrength (approved):

PR #2014, opened 2025-05-08
Fix: Rename 'hkey' to 'hKey' to match standard casing in similar APIs…

ajaymajukar:
… like RegOpenKeyExA

Ensures parameter naming consistency across registry API documentation.

### DIFF
--- a/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
+++ b/sdk-api-src/content/winreg/nf-winreg-reggetvaluea.md
@@ -63,7 +63,7 @@ Retrieves the type and data for the specified registry value.
 
 ## -parameters
 
-### -param hkey [in]
+### -param hKey [in]
 
 A handle to an open registry key. The key must have been opened with the KEY_QUERY_VALUE access right. For more information, see 
 <a href="/windows/desktop/SysInfo/registry-key-security-and-access-rights">Registry Key Security and Access Rights</a>. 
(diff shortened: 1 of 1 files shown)


HBelusca:
Does this match the casing used for the parameter of this function in the official Microsoft PSDK headers?

ajaymajukar:
Hi @HBelusca, thanks for addressing this PR.

As you asked, I have downloaded and installed the latest Windows SDK (version 10.0.26100) from the official site:
[https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/)

After checking the header file `winreg.h` at:
`C:\Program Files (x86)\Windows Kits\10.0.26100.0\um\winreg.h`

I noticed that **only the** `RegGetValueA` and `RegGetValueW` **functions use the parameter name** `hkey` **(lowercase ‘k’), while the rest of the related functions use** `hKey` **(camelCase) for their parameters**.

Since you asked whether the casing in the docs matches the official Microsoft PSDK headers for this specific function, I see that it does. Because of that, I think you would prefer to keep it consistent here in the MicrosoftDocs sdk-api as well.

However, I’m happy to defer to your guidance on how best to proceed with the pull request. Please let me know if you’d like me to update the PR accordingly or if you’d prefer to take it from here.

Thanks again for your help!

HBelusca:
I'm not from Microsoft, so I would suggest you wait for their reply instead. But from what I've noticed so far, their docs usually match what they have in their official headers.

drewbatgit:
Closing this PR — verification against the SDK headers shows that the official header uses lowercase `hkey`, not `hKey` as proposed. The existing documentation matches the actual header definition.

drewbatgit:
Rejected: SDK header verification confirms the existing lowercase \hkey\ parameter name is correct.

PR #2015, opened 2025-05-09
RmStartSession: Add information about strSessionKey size

oold:
Source for the previously missing information is https://devblogs.microsoft.com/oldnewthing/20120217-00/?p=8283.


drewbatgit:
@oold  Thanks for catching that from Raymond's blog post!

PR #2016, opened 2025-05-09
Fix error message in RemoveBindLink example

AlexCraciun6:


alvinashcraft (approved):
Looks good. Thanks for catching this copy/paste mistake.

PR #2019, opened 2025-05-15
Update nf-pathcch-pathcchstripprefix.md: Fix escape sequence

oold:
The description currently mentions a `\?` prefix. It should actually be the extended path prefix `\\?\`.


drewbatgit (approved):

PR #2025, opened 2025-06-01
Update `CreateColorProfileFromLogColorSpace` to steer people away from a memory leak

rickbrew:
The documentation indicates that you should use `GlobalFree` on the value stored at the address given by `pProfile`.

In other words, 

```c
BYTE* pProfile = NULL;
CreateProfileFromLogColorSpaceW(..., &pProfile);
GlobalFree((HGLOBAL)pProfile);
```

This is wrong, it is a memory leak. If you do this 65,536 times in a row then this method will start returning `ERROR_NOT_ENOUGH_MEMORY`. Apparently that's the maximum number of `HGLOBAL`s you can have.

The correct pattern is this:

```c
BYTE* pProfile = NULL;
CreateProfileFromLogColorSpaceW(..., &pProfile);
HGLOBAL hProfile = GlobalHandle(pProfile);
GlobalFree(hProfile);
```

I ran into this issue while working on my desktop app. I had a bug in a caching layer that was causing my color profile caching to not work, so it was recreating the profile on every call into the cache. After using the app for a few minutes it would crash because I convert Win32 errors into .NET exceptions.

cc @Sergio0694

### DIFF
--- a/sdk-api-src/content/icm/nf-icm-createprofilefromlogcolorspacea.md
+++ b/sdk-api-src/content/icm/nf-icm-createprofilefromlogcolorspacea.md
@@ -65,7 +65,7 @@ If the **lcsFilename** \[0\] member if the [**LOGCOLORSPACEA**](/windows/desktop
 
 ## -remarks
 
-This function can be used with ASCII or Unicode strings. The buffer created by this function must be freed by the caller when it is no longer needed or there will be a memory leak. The [GlobalFree](/windows/win32/api/winbase/nf-winbase-globalfree) function should be used to free this buffer.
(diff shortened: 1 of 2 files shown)


HBelusca:
Note that the "wrong" case you mention, could be correct only if internally, they were using `GlobalAlloc` with the `GMEM_FIXED` flag. (See also https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-globallock#remarks ).
Your observation of the memory leak, if one doesn't use the `GlobalLock` call (as your solution suggests), seems to indicate they didn't use the `GMEM_FIXED` flag... This is my understanding, at least.

drewbatgit (approved):

rickbrew:
Thanks @drewbatgit !

PR #2029, opened 2025-06-07
Update nf-dxgi-idxgiswapchain-getbuffer.md

ModulePillow:
The documentation indicates DXGI_SWAP_EFFECT_SEQUENTIAL and DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL have the same index mechanism, but it's incorrect.

The flip sequential model has a constant and consecutive index range according to the descriptions in learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect and my verification. Therefore, the paragraph related to this should be revised.


stevewhims:
Thank you for your research!

stevewhims (approved):

PR #2031, opened 2025-06-13
Fix incorrect ole32 documentation for COM DeviceCatalog APIs

brialmsft:
These APIs have always been implemented by combase.dll, not ole32.dll. They've never been in the ole32 import lib.

However, the combaseapi apiset had incorrect reverse forwarder configuration that identified ole32 as the legacy host of _all_ contained APIs, even those that were added to combase.dll after it was split from ole32.dll. The bad reverse forwarders resulted in bad import table entries for these APIs; I fixed that last year under bug 48441982.

I'm guessing that the incorrect documentation info also originated from that incorrect reverse forwarder.

### DIFF
--- a/sdk-api-src/content/combaseapi/nf-combaseapi-coregisterdevicecatalog.md
+++ b/sdk-api-src/content/combaseapi/nf-combaseapi-coregisterdevicecatalog.md
@@ -21,8 +21,8 @@ req.max-support:
 req.namespace: 
 req.assembly: 
 req.type-library: 
-req.lib: Ole32.lib
-req.dll: Ole32.dll
+req.lib: onecore.lib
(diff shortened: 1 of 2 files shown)


stevewhims (approved):

stevewhims (approved):

brialmsft:
@stevewhims just to sanity-check: I can't complete the PR myself, and somebody else will do it, right?

stevewhims:
> @stevewhims just to sanity-check: I can't complete the PR myself, and somebody else will do it, right?

Sorry about this! Somebody has removed my perms (and the perms of other writers) to Complete PRs in this repo (as I discovered after I approved it). I'll try and get them back.

PR #2036, opened 2025-06-20
WLX_PROFILE_V2_0: Tell how the string pointers are allocated

HBelusca:
The numerous `psz*` strings need to be allocated with LocalAlloc, since, according to investigations on Windows <= 2003, winlogon.exe will free these pointers with LocalFree.


alvinashcraft:
@GrantMeStrength These three WLX* PRs need some investigation or communication with the product team before merging the changes.

drewbatgit (approved):

PR #2040, opened 2025-06-25
Update nf-wingdi-createfontw.md

DJm00n:
Format legacy `*_CHARSET` constant tables.

### DIFF
--- a/sdk-api-src/content/dimm/ns-dimm-logfonta.md
+++ b/sdk-api-src/content/dimm/ns-dimm-logfonta.md
@@ -226,82 +226,27 @@ Type: <b>BYTE</b>
 
 Specifies the character set. The following values are predefined:
 
-<table class="clsStd">
-<tr>
-<td>ANSI_CHARSET</td>
-<td>BALTIC_CHARSET</td>
-</tr>
-<tr>
-<td>CHINESEBIG5_CHARSET</td>
-<td>DEFAULT_CHARSET</td>
-</tr>
-<tr>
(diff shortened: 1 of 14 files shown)


stevewhims:
Thanks, @DJm00n, I like the idea in general of changing HTML to markdown!

But why are we making this particular change? Note that we're losing value by removing headings such as "Korean Windows". Is the expectation that readers will already know that JOHAB_CHARSET == Korean? Or if not, they can go look it up? Seems dubious.

In any case, for a constant such as JOHAB_CHARSET, I'd put that in bold (asterisk asterisk), and not backticks. Backticks are for code, filenames, that kind of thing.

Also, parameter names should be in italics (as they were before this change), not backticks (as they are after this change). But again changing HTML to markdown italics is desirable.

So I won't take this PR right now.

Thanks! :)
-Steve

DJm00n:
@stevewhims fixed formatting as you're asked.

>Note that we're losing value by removing headings such as "Korean Windows". Is the expectation that readers will already know that JOHAB_CHARSET == Korean? Or if not, they can go look it up? Seems dubious.

I removed headings such as "Korean Windows" because any Windows version since NT have these defines. These are legacy font codepages that are there for backwards compatibility.

Actual table was borrowed here: https://github.com/MicrosoftDocs/windows-driver-docs-ddi/blame/staging/wdk-ddi-src/content/prntfont/ns-prntfont-_printifi32.md#L109-L129

Ref: https://github.com/wine-mirror/wine/blob/master/dlls/win32u/font.c#L291-L311

PR #2045, opened 2025-07-06
Update nf-dxgi1_3-createdxgifactory2.md

kryptoid256:
Fix the vocabulary


stevewhims (approved):
I'm not sure changing "object" to "interface" was important here. However, I've made a couple changes that I do think help a little. I didn't change all HTML to markdown, but doing that would improve the topic, too. :)

PR #2052, opened 2025-07-18
Update nf-processenv-setenvironmentstringsw.md

Wolf-SO:
Fix errors in format and return value specification.

For error in return value description, see:
https://github.com/MicrosoftDocs/feedback/issues/2235

The correct format of the environment variable block is already described in the `GetEnvironmentStrings` function(s).


Wolf-SO:

Karl-Bridge-Microsoft (approved):

PR #2055, opened 2025-07-21
Fix ID3D11Multithread docs: use D3d11.lib/dll instead of D3d11_4.lib/dll

fatpound:
The docs incorrectly listed D3d11_4.lib and D3d11_4.dll as dependencies for ID3D11Multithread, but I could not find them anywhere.
I also looked into _1, _2, and _3 versions and couldn't find separate lib or DLL files for those libraries either.
After checking SDK folders and online sources, it looks like the correct ones are D3d11.lib and D3d11.dll

This update corrects those references to prevent confusion and build errors for developers.


fatpound:

stevewhims (approved):

PR #2059, opened 2025-07-30
Update nf-winenclaveapi-enclaverestrictcontainingprocessaccess.md

ericlaw1979:
Clarify what happens when access is restricted, and remove implication that the EnclaveCopy* APIs only work when access was restricted via the EnclaveRestrictContainingProcessAccess API (they also work when image_enclave_policy_strict_memory is set).


alvinashcraft (approved):
Thanks for this wording change. This looks good.

PR #2061, opened 2025-08-01
Fix new file flags

0xbadfca11:
- Only the disallow flag is defined and is disabled by default.
- From the Windows SDK, this definition is incorrect.


sdk-api-src/content/fileapi/nf-fileapi-deletefile2a.md
```diff
@@ -68,7 +68,7 @@ Flags to specify how to treat the file that is being deleted. This parameter can
 
 | Value | Meaning |
 |-------|-------------|
-| **FILE_FLAGS_DISALLOW_PATH_REDIRECTS**<br/>`0x00000001` | Prevent *lpFileName* from being redirected by reparse points or symbolic links. |
+| **FILE_FLAG_DISALLOW_PATH_REDIRECTS**<br/>`0x00010000` | Prevent *lpFileName* from being redirected by reparse points or symbolic links. |
```
alvinashcraft, line 71: Thanks! I confirmed this updated name and value is correct in winbase.h.

sdk-api-src/content/fileapi/nf-fileapi-createfile3.md
```diff
@@ -133,7 +133,7 @@ If the function fails, the return value is **INVALID_HANDLE_VALUE**. To get exte
 
 ## -remarks
 
-**CreateFile3** behaves exactly the same way as [CreateFile2](nf-fileapi-createfile2.md) with one exception; operations will fail if *lpFileName* is redirected via a reparse point or symbolic link. This behavior can be disabled with a new flag that can be added to the *dwFileFlags*.
+**CreateFile3** behaves exactly the same way as [CreateFile2](nf-fileapi-createfile2.md) with one exception; operations can fail if *lpFileName* is redirected via a reparse point or symbolic link. This behavior can be enabled with a new flag that can be added to the *dwFileFlags*.
```
alvinashcraft: Would this be clearer, rather than referring to the behavior as "enabled" or "disabled":
```suggestion
**CreateFile3** behaves exactly the same way as [CreateFile2](nf-fileapi-createfile2.md) with one exception; operations can fail if *lpFileName* is redirected via a reparse point or symbolic link. Redirects can be disallowed by adding the **FILE_FLAG_DISALLOW_PATH_REDIRECTS** flag to the *dwFileFlags*.
```

alvinashcraft (commented):
Added a suggestion inline. Otherwise, this fix looks good. Thanks for your contribution!

alvinashcraft (approved):

PR #2065, opened 2025-08-04
Add DST ambiguity note to TzSpecificLocalTimeToSystemTime(Ex) documentation

rkg96:
This PR updates the Remarks section of the [TzSpecificLocalTimeToSystemTime](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-tzspecificlocaltimetosystemtime) and [TzSpecificLocalTimeToSystemTimeEx](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-tzspecificlocaltimetosystemtimeex) APIs documentation to clarify behavior around Daylight Saving Time (DST) transitions.

Specifically, it adds a note explaining that:
- Certain local times near DST transitions may be ambiguous (e.g., repeated during fall back) or invalid (e.g., skipped during spring forward).
- The API may not return the "correct" UTC value in these cases due to the lack of a DST indicator in the SYSTEMTIME structure.
- Developers should be aware of this limitation and consider alternative approaches for time-sensitive operations.

This clarification is based on MSFT-internal bug reports and developer feedback and aims to prevent confusion and misinterpretation of API behavior during DST transitions.


rkg96:
@microsoft-github-policy-service agree company="Microsoft"

sdk-api-src/content/timezoneapi/nf-timezoneapi-tzspecificlocaltimetosystemtime.md
```diff
@@ -88,6 +88,13 @@ If the function fails, the return value is zero. To get extended error informati
 
 <b>TzSpecificLocalTimeToSystemTime</b> takes into account whether daylight saving time (DST) is in effect for the local time to be converted.
 
+> [!IMPORTANT]
+> Certain local times near DST transitions may be <b>ambiguous</b> or <b>invalid</b>. Specifying such a time can result in unexpected behavior as there is no guaranteed "correct" result. 
+> - During the transition from daylight time to standard time, the local clock repeats. A local time within the repeated window is ambiguous because it occurs twice, once in daylight time and once in standard time.
+> - During the transition from standard time to daylight time, the local clock jumps forward. A local time within the skipped window does not have a valid UTC conversion. 
+>
+> Applications requiring continuity or precision should avoid relying on this function and look to use UTC time directly instead.
```
Karl-Bridge-Microsoft: This final paragraph should be the initial statement for the note. It gets lost here and requires the reader to read all the preamble, which should instead just be the explanation for the opening statement. It also isn't very firm in its guidance. Do any apps not require continuity or precision? Is there a cost to using UTC directly. Is this function effectively deprecated and not recommended at all?

Same recommendation for the other "ex" topic.

Karl-Bridge-Microsoft (approved):

PR #2067, opened 2025-08-09
Clarify that pwszObjectName needs to be null terminated in WinHttpOpenRequest

maj113:
This better matches the Documentation of [WinHttpConnect](https://learn.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpconnect#parameters) which does specify that the host name string must be null terminated


stevewhims:
Ok. Seems very low pri, since the parm name indicates it's null-terminated, there's no way to provide the string's length, and it's hard to imagine someone providing a string that's not null-terminated these days.

PR #2070, opened 2025-08-13
Update ns-winternl-peb_ldr_data.md

ImJustLegit:
Updated the LDR_DATA_TABLE_ENTRY struct definition to match the one in Windows version 24H2.


guettafa:

PR #2080, opened 2025-09-01
Update nf-shlobj_core-shgetknownfolderitem.md

Aemony:
Removed a sentence that didn't make sense in the context. The caller does not *always* need administrator privileges -- it only needs it when making a call on public known folders.


drewbatgit (approved):

PR #2082, opened 2025-09-03
Update ne-wtsapi32-wts_info_class.md to include WTSSessionActivityId …

kehedgesMsft:
Include WTSSessionActivityId description in Constants table


mattwojo:
Moved this update to the private repo and rolling it out to the live site now.

PR #2086, opened 2025-09-08
CheckMenuItem article update

nikita-leontiev:
CheckMenuItem returns UINT_MAX in case of failure.


drewbatgit (approved):

PR #2094, opened 2025-09-23
Fix invalid UTF-8 encoding

m417z:
```
--------------------------------------------------
Hex    Count    Percentage   Character
--------------------------------------------------
a0     1648       98.6%      ' '
96     18          1.1%      '–'
ae     4           0.2%      '®'
97     2           0.1%      '—'
--------------------------------------------------
```

Note that 85 files also contain the Replacement Character (�). These files weren't fixed, they have to be fixed manually since the character can't be recovered automatically.
**Edit**: See #2095.


GrantMeStrength:
All these changes need to be confirmed manually, and there are so many the tool goes too slowly. It's unlikely I will have time to approve these.

m417z:
Note that this was an automated replacement. If you prefer to verify it, you can create a script which simply replaces the specified bytes and verify the result.

GrantMeStrength:
Every correct to the docs needs to be manually verified by policy. We can't just accept bulk updates.

m417z:
I understand. Let me know if there's anything else I can do to help with this, like providing a Python script which does the replacements.

GrantMeStrength:
Is there a way to focus on the most egregious errors, and then break them up into smaller PRs?
I appreciate your time and effort BTW!

m417z:
I don't think there are more or less egregious errors here. There are just many files which aren't valid UTF-8, which is the format everyone expects nowadays. I don't think going over manually is the best way to handle it, but if you have to and if it will help, I can split it to chunks and open several PRs.

drewbatgit (approved):

PR #2095, opened 2025-09-23
Fix invalid UTF-8 characters

m417z:
The fixes were made by Claude Sonnet 4


ChrisDenton:
Just very quickly skimming this but a lot of these seem wrong. `�` is being erroneously replaced with `-` instead of ` `.

m417z:
Thanks, I made another pass that replaces some of the dashes with spaces. Let me know if you still find anything that's incorrect.

sdk-api-src/content/control/nf-control-imediacontrol-getstate.md
```diff
@@ -1,7 +1,7 @@
 ---
 UID: NF:control.IMediaControl.GetState
 title: IMediaControl::GetState (control.h)
-description: The GetState method retrieves the state of the filter graph—paused, running, or stopped.
+description: The GetState method retrieves the state of the filter graph: paused, running, or stopped.
```
m417z: @GrantMeStrength please note:
1. It's not valid to use `:` in a string in YAML without quotes.
2. That's the text the documentation itself contains:
   https://github.com/MicrosoftDocs/sdk-api/blob/3d873fdaf15d3f07be52bf8259f808428c931da8/sdk-api-src/content/control/nf-control-imediacontrol-getstate.md#L56

It's also the case for some of your other changes - the dash can also be found in the documentation itself.

m417z:
Unfortunately some of the follow-up commits that were merged caused some of the files to become invalid due to the `:` symbol, for example:
https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/msrdc/nf-msrdc-irdcgeneratorfiltermaxparameters-sethashwindowsize.md

PR #2097, opened 2025-09-24
Escape Markdown formatting

ihor-kitsa:
There are asterisks in text that Markdown replaces with \<em> HTML tags. The first asterisk needs an escape character.


drewbatgit (approved):

PR #2110, opened 2025-10-06
Update nf-winevt-evtgetpublishermetadataproperty.md

iPhilip:
When the function fails, it always fails with **FALSE**.


iPhilip:
While it fails with FALSE, the [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror) function returns with ERROR_INSUFFICIENT_BUFFER so the text is ok.

PR #2116, opened 2025-10-15
Button_* macro articles update

nikita-leontiev:
Return values corrected to match the SDK.


drewbatgit (approved):

PR #2118, opened 2025-10-16
Update raytracing acceleration structure requirements

Jasper-Bekkers:
Added requirement for D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS to be set when using D3D12_RESOURCE_FLAG_RAYTRACING_ACCELERATION_STRUCTURE.

### DIFF
--- a/sdk-api-src/content/d3d12/ne-d3d12-d3d12_resource_flags.md
+++ b/sdk-api-src/content/d3d12/ne-d3d12-d3d12_resource_flags.md
@@ -138,7 +138,7 @@ Specfies that this resource may be used only as an encode reference frame. It ma
 
 ### -field D3D12_RESOURCE_FLAG_RAYTRACING_ACCELERATION_STRUCTURE:0x100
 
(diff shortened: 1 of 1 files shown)


jenatali (approved):

MarijnS95:
Note that this is documented at https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#acceleration-structure-memory-restrictions:

> Further, resources that will contain acceleration structures must be created in the state [`D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE`](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#additional-resource-states), and must have resource flag `D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS`. The `ALLOW_UNORDERED_ACCESS` requirement simply acknowledges both: that the system will be doing this type of access in its implementation of acceleration structure builds behind the scenes, and from the app point of view, synchronization of writes/reads to acceleration structures is accomplished using UAV barriers (discussed later).

For the validation message, there's an errant `_VIEW` suffix (making it harder to copy+search), CC @jenatali:

```
[ID: D3D12_MESSAGE_ID(599)] - [STATE_CREATION ERROR] "ID3D12Device::CreatePlacedResource2: Resource created with D3D12_RESOURCE_FLAG_RAYTRACING_ACCELERATION_STRUCTURE bit set must also set the D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS_VIEW bit."
```

`D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS_VIEW` doesn't exist.

stevewhims (approved):

PR #2121, opened 2025-10-20
Fix some typos and bad info in fileapi.h docs

majaha:
These functions' docs referred to parameter names that didn't exist.

Also, the docs for the [`GetLogicalDriveStringsW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlogicaldrivestringsw) function's `nBufferLength` parameter said "This size does not include the terminating null character" but this isn't true as far as I can tell.

### DIFF
--- a/sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
@@ -70,9 +70,8 @@ Fills a buffer with strings that specify valid drives in the system.
 
 ### -param nBufferLength [in]
 
-The maximum size of the buffer pointed to by <i>lpBuffer</i>, in 
-      <b>TCHARs</b>. This size does not include the terminating null character. If this 
-      parameter is zero, <i>lpBuffer</i> is not used.
+The maximum size of the buffer pointed to by <i>lpBuffer</i>, in
+      <b>TCHARs</b>. If this parameter is zero, <i>lpBuffer</i> is not used.
 
 ### -param lpBuffer [out]
 
--- a/sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
+++ b/sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
@@ -70,8 +70,8 @@ Fills a buffer with strings that specify valid drives in the system.
 
 ### -param nBufferLength [in]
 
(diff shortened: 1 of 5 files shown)


majaha:

alvinashcraft:
Thanks for these updates. I'll verify this part with the developers before merging the changes: "This size does not include the terminating null character".

sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
```diff
@@ -70,9 +70,8 @@ Fills a buffer with strings that specify valid drives in the system.
 
 ### -param nBufferLength [in]
 
-The maximum size of the buffer pointed to by <i>lpBuffer</i>, in
```
alvinashcraft, line 73: @majaha here's what I heard back from the engineering folks:

> I do think it requires the parameter to include the null byte. 
>
> For instance, if I call the API with 0 and null as arguments, I get back that the required number of characters is 5 (if I have a single drive), even though the max is 4 characters, so this includes the null byte. If I then pass in 4 as the size argument, it fails at filling the buffer and returns 5, but if I pass in 5, it succeeds at filling the buffer and returns 4.  
>
> As far as I can tell, this API is just implemented in a wonky way. Anyways, they need to include the null byte in the size parameter.

alvinashcraft (changes requested):
I'm updating this to require some changes based on the comment from engineering about nBufferLength. Thanks!

sdk-api-src/content/fileapi/nf-fileapi-getlogicaldrivestringsw.md
majaha: Yes, what the engineer is describing matches the behaviour I was seeing. The return value is a little unusual, and unrelatedly, the null byte needs to be included in the size parameter.
So, removing the language `This size does not include the terminating null character.` from the documentation of the size parameter is correct. Are we in agreement on that?

alvinashcraft: I don't know if it should be removed or changed to `This size should include the terminating null character'.

majaha: Okay, I've added some text explaining that the parameter includes space for the terminating null character. How's that?

alvinashcraft (approved):
This looks good to me. Thanks for the fixes and for working through the odd buffer size issue.

PR #2123, opened 2025-10-23
Change DirectX 12 Agility SDK requirement to 1.6

JackiBackiBoy:
Updated the required version of the DirectX 12 Agility SDK for Enhanced Barriers and Relaxed Format Casting support to version 1.6. It was previously listed that 1.7 was required, but no such Agility SDK version exists and is incorrect.


JackiBackiBoy:
I forgot to mention this in the pull request, but technically SDK version 1.6 and above is not really correct, it should be Agility SDK version 1.608.0 and above. This is confirmed in the changelog of 1.608.0.
URL: https://www.nuget.org/packages/Microsoft.Direct3D.D3D12/1.608.0

stevewhims (approved):

PR #2131, opened 2025-11-03
Fix ambiguous text of GetCopyableFootprints function

alpakaDurumi:
The current explanation for `BaseOffset` parameter is ambiguous. It is described as `The offset, in bytes, to the resource.`, which can be incorrectly interpreted as an offset of the resource represented by the `pResourceDesc` parameter.

In the function, `BaseOffset` is not an offset for the resource. Instead, it is a value that is simply added to the `Offset` of each subresource layout in the output `pLayouts` array.

For reference, see `d3dx12.h` version of `GetCopyableFootprints` : https://github.com/microsoft/DirectX-Headers/blob/18d5bfe96452667b201fd26aa25c6c305ae57e07/include/directx/d3dx12_resource_helpers.h#L520


alpakaDurumi:

stevewhims (approved):

PR #2150, opened 2025-12-12
IExtractImage added to the interface list

nikita-leontiev:


drewbatgit (approved):

PR #2153, opened 2025-12-16
Update nf-winuser-getmenuiteminfoa.md

iPhilip:
Changes _uItem_ to _Item_ for consistency with the function's syntax and the function's description in the header file.


drewbatgit (approved):

PR #2158, opened 2025-12-19
Update ns-ntsecapi-kerb_external_name.md

WAftring:
Fix typo in text


alvinashcraft (approved):
Thanks for finding and fixing this. I also just double checked that the typo wasn't also in the ntsecapi header, just to be sure. This is good to go.

PR #2161, opened 2026-01-01
Fix "returns a pointer" for out parameters

pfusik:


pfusik:
Please check if there are more of these unfortunate wordings.

PR #2162, opened 2026-01-02
Escape Markdown characters

Meterel:
Escape Markdown characters like it's done in https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-gettokeninformation in pages like https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocesstoken


Meterel:

PR #2192, opened 2026-03-30
Update documentation for IMAGE_FILE_MACHINE_ARM64

anporumb:
Added ARM64 architecture support details to documentation.


anporumb:
I take it back - StackWalk's support for ARM64 is... questionable. Might work on Ampere Altra, does not work on Cobalt as intended and there needs to be some additional work until it does. So I will close this PR which I have no fully vetted before asking for it.

PR #2202, opened 2026-04-23
Update remarks section to use wcslen instead of strlen

RatinCN:
`BSTR` contains `WCHAR`s, you probably mean `wcslen` here.

### DIFF
--- a/sdk-api-src/content/oleauto/nf-oleauto-sysstringlen.md
+++ b/sdk-api-src/content/oleauto/nf-oleauto-sysstringlen.md
@@ -64,7 +64,7 @@ The number of characters in <i>bstr</i>, not including the terminating null char
 
 ## -remarks
 
-The returned value may be different from <b>strlen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
+The returned value may be different from <b>wcslen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
 
 ## -see-also
 


sdk-api-src/content/oleauto/nf-oleauto-sysstringlen.md
```diff
@@ -64,8 +64,8 @@ The number of characters in <i>bstr</i>, not including the terminating null char
 
 ## -remarks
 
-The returned value may be different from <b>strlen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
+The returned value may be different from <b>wcslen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
```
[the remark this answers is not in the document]
GrantMeStrength: Accepted — you're right, the parameter is formally named `pbstr` and the remarks should be consistent with that. Worth noting the `-returns` section has the same pre-existing inconsistency (`bstr` appears there too), so I'll fix both in the same pass.

sdk-api-src/content/oleauto/nf-oleauto-sysstringlen.md
```diff
@@ -64,8 +64,8 @@ The number of characters in <i>bstr</i>, not including the terminating null char
 
 ## -remarks
```
GrantMeStrength, line 65: The `-returns` section has the same inconsistency — `bstr` should be `pbstr` to match the parameter name.

```suggestion
The number of characters in <i>pbstr</i>, not including the terminating null character. If <i>pbstr</i> is null the return value is zero.
```

sdk-api-src/content/oleauto/nf-oleauto-sysstringlen.md
```diff
@@ -64,8 +64,8 @@ The number of characters in <i>bstr</i>, not including the terminating null char
...
 
-The returned value may be different from <b>strlen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
+The returned value may be different from <b>wcslen</b>(bstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
 
 ## -see-also
```
GrantMeStrength, line 69: Applying the Copilot suggestion — `bstr` → `pbstr` to match the parameter name.

```suggestion
The returned value may be different from <b>wcslen</b>(pbstr) if the BSTR contains embedded Null characters. This function always returns the number of characters specified in the cch parameter of the <a href="/previous-versions/windows/desktop/api/oleauto/nf-oleauto-sysallocstringlen">SysAllocStringLen</a> function used to allocate the BSTR.
```

PR #2216, opened 2026-05-26
Update SDK API docs with changes for ML-DSA and ML-KEM in Bcrypt

jess-krynitsky:
Brought over doc updates from Win32 that were updated last year and missed from the SDK docs.

### DIFF
--- a/sdk-api-src/content/bcrypt/nf-bcrypt-bcryptdecapsulate.md
+++ b/sdk-api-src/content/bcrypt/nf-bcrypt-bcryptdecapsulate.md
@@ -47,9 +47,6 @@ api_name:
 
 # BCryptDecapsulate function
 
-> [!NOTE]
-> Some information relates to a prerelease product which may be substantially modified before it's commercially released. Microsoft makes no warranties, express or implied, with respect to the information provided here. The feature described in this topic is available in pre-release versions of the [Windows Insider Preview](https://www.microsoft.com/software-download/windowsinsiderpreviewSDK).
-
 The **BCryptDecapsulate** function performs the Decapsulation operation of a Key Encapsulation Mechanism (KEM).
 It takes a KEM ciphertext and it decrypts with the provided private key, returning the shared secret key.
 
--- a/sdk-api-src/content/bcrypt/nf-bcrypt-bcryptdecapsulate.md
+++ b/sdk-api-src/content/bcrypt/nf-bcrypt-bcryptdecapsulate.md
@@ -1,0 +1,138 @@
+---
+UID: NF:bcrypt.BCryptDecapsulate
+title: BCryptDecapsulate function (bcrypt.h)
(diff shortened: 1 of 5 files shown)


GrantMeStrength (commented):
@JessicaKryworking — a few issues I spotted in this PR:

1. **`BCryptEncapsulate` title contains website chrome** — The H1 heading reads `# BCryptEncapsulate function - Win32 apps | Microsoft Learn`. It should just be `# BCryptEncapsulate function`. Same issue on `BCRYPT_PQDSA_KEY_BLOB` (`# BCRYPT_PQDSA_KEY_BLOB structure - Win32 apps | Microsoft Learn`).

2. **`BCryptEncapsulate` uses wrong callout syntax** — The NOTE is rendered as plain text (`Note`) instead of using the Learn callout syntax (`> [!NOTE]`). The `BCryptDecapsulate` file gets this right but `BCryptEncapsulate` and `BCRYPT_PQDSA_KEY_BLOB` do not.

3. **Contradictory prerelease disclaimers** — Several files include the "prerelease product" disclaimer, but `req.target-min-winverclnt` is set to `Windows 11 24H2` which is already released. The disclaimers should either be removed or the min version should reflect the actual preview build.

4. **`original_content_git_url` points to `win32-pr`** — The YAML front-matter in `BCryptDecapsulate` and `BCryptEncapsulate` has `original_content_git_url` pointing to `win32-pr` rather than `sdk-api`. This may cause tooling confusion.

5. **Missing composite algorithm content** — The `BCRYPT_MLKEM_KEY_BLOB` and `BCRYPT_PQDSA_KEY_BLOB` files here are copies of the older win32 versions and don't include the composite algorithm updates from win32 PR #2186. Once merged, these sdk-api copies will be immediately stale. Consider incorporating the composite struct/magic/parameter-set additions here too.

6. **Missing trailing newline** — A few files are missing a final newline character.

GrantMeStrength (commented):
@jess-krynitsky — (correcting the earlier mention) — please see my review comments above regarding title chrome, callout syntax, prerelease disclaimers, original_content_git_url, missing composite content, and trailing newlines.

jess-krynitsky:
@GrantMeStrength just pushed some updates:
1. Updated headers to remove "Win32" reference
2. Fixed Note callout syntax on all files
3. For the prerelease disclaimer - this applies only to the composite algorithm updates in the key blob pages. I removed the note from the encapsulate and decapsulate pages, and on the key blob pages I updated the wording to explicitly callout the composite features as in prerelease. The general ML-KEM and PQDSA key blobs are publicly released.
4. I just removed the original_content_git_url from the metadata since these are new files in the SDK API and don't want any tooling problems.
5. Updated key blob files with composite content to match Win32 PR updates.
6. Added trailing newline character to the files that were missing it.

Thank you!

PR #2218, opened 2026-05-30
Fix inconsistency in PathCchIsRoot documentation

csware:
The remarks after the table were inconsistent with the table. I checked what result actually holds and updated the example.

### DIFF
--- a/sdk-api-src/content/pathcch/nf-pathcch-pathcchisroot.md
+++ b/sdk-api-src/content/pathcch/nf-pathcch-pathcchisroot.md
@@ -177,4 +177,4 @@ The following table shows the <b>PathCchIsRoot</b> return value for various path
 </table>
  
 
-This function returns <b>TRUE</b> for paths such as "\", "<i>X</i>:\" or "&#92;&#92;<i>server</i>&#92;<i>share</i>". Paths such as "..\path2" or "&#92;&#92;<i>server</i>\" return <b>FALSE</b>.
+This function returns <b>TRUE</b> for paths such as "\", "<i>X</i>:\" or "&#92;&#92;<i>server</i>&#92;<i>share</i>". Paths such as "..\path2" or "&#92;&#92;<i>server</i>\share\" return <b>FALSE</b>.


sdk-api-src/content/pathcch/nf-pathcch-pathcchisroot.md
```diff
@@ -177,4 +177,4 @@ The following table shows the <b>PathCchIsRoot</b> return value for various path
 </table>
  
 
-This function returns <b>TRUE</b> for paths such as "\", "<i>X</i>:\" or "&#92;&#92;<i>server</i>&#92;<i>share</i>". Paths such as "..\path2" or "&#92;&#92;<i>server</i>\" return <b>FALSE</b>.
+This function returns <b>TRUE</b> for paths such as "\", "<i>X</i>:\" or "&#92;&#92;<i>server</i>&#92;<i>share</i>". Paths such as "..\path2" or "&#92;&#92;<i>server</i>\share\" return <b>FALSE</b>.
```
GrantMeStrength, line 180: Good fix! Small formatting nit per the Copilot review — the UNC path in the FALSE example should use `&#92;` escapes and `<i>` tags to match the rest of the line. Suggested fix:

```suggestion
This function returns <b>TRUE</b> for paths such as "\", "<i>X</i>:\" or "&#92;&#92;<i>server</i>&#92;<i>share</i>". Paths such as "..\path2" or "&#92;&#92;<i>server</i>&#92;<i>share</i>\" return <b>FALSE</b>.
```

GrantMeStrength (approved):
The change is correct and an improvement in clarity.

**Verification against the table:**
- `\\path1\path2` → TRUE (= `\\server\share`) ✅ — stays in the TRUE list
- `\\path1\path2\` → FALSE (= `\\server\share\`) ✅ — the new FALSE example

The original prose wasn't wrong (`\\path1\` → FALSE per the table), but the new example is pedagogically clearer: it makes the TRUE/FALSE pair directly parallel so readers can see that adding a trailing backslash to a UNC root changes the result to FALSE. The old example compared paths that differed in *two* ways (share component and trailing slash); the new one isolates just the trailing slash.

PR #2222, opened 2026-06-15
winldap: Fix ldap_get_dn function prototype

jay:
- Override the syntax manually to declare ldap_get_dn returns PTCHAR.

Prior to this change the auto-generated syntax erroneously said that ldap_get_dn returned PCHAR instead of PTCHAR.

The bad syntax probably has to do with the way the unusual way ldap functions are declared in winldap.h. ldap_get_dn is either a macro that maps to ldap_get_dnW (LDAP_UNICODE defined, returns PWCHAR) or given its own protoype the same as ldap_get_dnA (no LDAP_UNICODE, returns PCHAR).

Closes #xxxx

---

AFAICT, LDAP_UNICODE is only used internally and set depending on UNICODE. They are different symbols so LDAP_UNICODE can be different from UNICODE. This presents a dilemma because LDAP_UNICODE does not have specific dependent types like TCHAR. It may be more correct (but also confusing) to add in remarks that if LDAP_UNICODE is defined to 0 then the return type is PCHAR. 

### **Many ldap functions have incorrect prototype documentation but I only changed one. There is no way to open an issue so I opened a PR with this one change to notify. I suggest MS rather tackle this issue wholesale.**

### DIFF
--- a/sdk-api-src/content/winldap/nf-winldap-ldap_get_dn.md
+++ b/sdk-api-src/content/winldap/nf-winldap-ldap_get_dn.md
@@ -50,6 +50,15 @@ api_name:
 # ldap_get_dn function
 
 
+## -syntax
+
+```cpp
+WINLDAPAPI PTCHAR LDAPAPI ldap_get_dn(
+  [in] LDAP        *ld,
+  [in] LDAPMessage *entry
+);
+```
+
 ## -description
 
 The <b>ldap_get_dn</b> function retrieves the distinguished name for a given entry.
--- a/sdk-api-src/content/winldap/nf-winldap-ldap_get_dn.md
+++ b/sdk-api-src/content/winldap/nf-winldap-ldap_get_dn.md
@@ -53,10 +53,23 @@ api_name:
 ## -syntax
 
 ```cpp
-WINLDAPAPI PTCHAR LDAPAPI ldap_get_dn(
+WINLDAPAPI PWCHAR LDAPAPI ldap_get_dnW(
   [in] LDAP        *ld,
   [in] LDAPMessage *entry
 );
+
+#if LDAP_UNICODE
+
+#define ldap_get_dn ldap_get_dnW
+
+#else
+
+WINLDAPAPI PCHAR LDAPAPI ldap_get_dn(
+  [in] LDAP        *ld,
+  [in] LDAPMessage *entry
+);
+
+#endif
 ```
 
 ## -description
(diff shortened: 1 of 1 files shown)


GrantMeStrength (approved):
LGTM — accurate technical correction. Approved.

jay:
Please note many ldap functions are affected by this and it may be better for MS to make some structural change to address it due to the unusual way the functions are declared, as mentioned in my original post.

GrantMeStrength:
@copilot /review

GrantMeStrength (approved):

GrantMeStrength (changes requested):
## Maintainer audit: systemic WinLDAP documentation issue

**Requesting changes because the proposed `PTCHAR` override is not correct for every supported preprocessor configuration. This supersedes my earlier approvals.**

`winldap.h` selects the generic API with `LDAP_UNICODE`, not directly with `UNICODE`:

```cpp
#ifndef LDAP_UNICODE
#ifdef UNICODE
#define LDAP_UNICODE 1
#else
#define LDAP_UNICODE 0
#endif
#endif

#if LDAP_UNICODE
#define ldap_get_dn ldap_get_dnW  // returns PWCHAR
#else
WINLDAPAPI PCHAR LDAPAPI ldap_get_dn(...);
#endif
```

`PTCHAR` follows `UNICODE`, so it can disagree with the function selected when a caller defines `LDAP_UNICODE` independently. The generic return type is conditionally `PWCHAR` or `PCHAR`; there is no existing TCHAR typedef governed by `LDAP_UNICODE`.

### Scope confirmed by repository/header audit

- **72 generic WinLDAP names are controlled by `LDAP_UNICODE`.**
- **62 have generic documentation topics in this repository.**
- **60 of those topics expose character-sensitive ANSI types** such as `PSTR`, `PCHAR`, `PZPSTR`, `LDAPControlA`, or `LDAPModA` even though the generic name maps to the Unicode entry point when `LDAP_UNICODE != 0`.
- Only `ldap_get_option` and `ldap_set_option` have encoding-neutral visible prototypes.
- The affected topics span connection, bind, search, modify, rename, add, compare, delete, result/control parsing, attribute/value/DN, sort, paging, reference, and extended-operation APIs.
- The generated A/W-topic warning also says selection is based on `UNICODE`; strictly, `winldap.h` selects using `LDAP_UNICODE` (which merely defaults from `UNICODE`).

This confirms the PR author's observation that the problem is systemic and caused by the unusual conditional declarations in `winldap.h`. I found no other open fix for the underlying harvesting behavior.

### Required direction

Please do not merge this as a standalone `PTCHAR` substitution. At minimum, this topic must document both branches explicitly (`PWCHAR` when `LDAP_UNICODE != 0`, `PCHAR` otherwise) and explain the override behavior. The durable solution should address the shared harvesting/template behavior for the affected WinLDAP generic topics rather than applying approximately 60 independent `PTCHAR` substitutions.

jay:
I wasn't sure the best way to explain, but I took a shot as you can see in the fixup commit I just pushed:

> The SDK sets <b>LDAP_UNICODE</b> to 1 or 0 depending on whether <b>UNICODE</b> is defined. You may override that behavior by setting it yourself.

I would encourage MS to pursue the "durable solution" but I don't know how to do that.

PR #2231, opened 2026-07-03
Enhance documentation for SetThreadCursorCreationScaling

DJm00n:
Updated the documentation for SetThreadCursorCreationScaling function to include detailed parameters, return values, and remarks regarding DPI scaling behavior.

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-setthreadcursorcreationscaling.md
+++ b/sdk-api-src/content/winuser/nf-winuser-setthreadcursorcreationscaling.md
@@ -43,7 +43,7 @@ dev_langs:
 
 ## -description
 
-Sets the DPI scale for which the cursors being created on this thread are intended. This value is taken into account when scaling the cursor for the specific monitor on which it is being shown.
+Sets the DPI scale for which the cursors being created on this thread are intended. This value is taken into account when scaling the cursor based on the current system DPI.
 
 ## -parameters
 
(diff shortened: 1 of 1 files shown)


GrantMeStrength (changes requested):
The expanded remarks are a big improvement, but two issues need addressing:

**1. Description change introduces an inconsistency**

The description is changed from:
> "when scaling the cursor for the specific monitor on which it is being shown"

to:
> "when scaling the cursor based on the current system DPI"

But the companion PR (win32/2196) refers to this same function as enabling "automatic **per-monitor** DPI scaling". If the function scales per-monitor (which the original wording suggested), the new description is a regression. If it genuinely only tracks system DPI, then the companion PR is wrong. Please clarify which is accurate and make both PRs consistent.

**2. Placeholder function in code example**

`CreateCursorFrom48pxData(...)` is not a real API. Either replace with a real function call (e.g. `CreateCursor`) or add a comment making clear this is a stand-in:

```cpp
// Create a 48x48 cursor tagged for 144 DPI (150% scale)
UINT previousDpi = SetThreadCursorCreationScaling(144);
HCURSOR hCursor = CreateCursor(/* ... your cursor data ... */);
SetThreadCursorCreationScaling(previousDpi);
```

GrantMeStrength:
@copilot /review

GrantMeStrength (dismissed):
**Review: Approved ✅**

Good documentation addition for SetThreadCursorCreationScaling. The code sample correctly demonstrates the save/set/create/restore pattern, and the explanation of when this API is needed (programmatic cursor creation vs. resource-loaded cursors) is clear and valuable.

Minor note: `CreateCursorFrom48pxData(...)` is clearly a placeholder — this is fine for illustrating the pattern.

GrantMeStrength:
thanks!

PR #750, opened 2021-04-26
Document observed behavior of possible return values

Youw:
If E_POINTER is not explicitly documented as an expected check,
one should make an explicit check in `QueryInterface` implementation, beforce calling `QISearch`.

PR #2234, opened 2026-07-13
Document observed behavior of possible return values

Youw:
If E_POINTER is not explicitly documented as an expected check,
one should make an explicit check in `QueryInterface` implementation, beforce calling `QISearch`.


Youw:
Still relevant even after #750 is closed.

GrantMeStrength (approved):
Correct addition — `QISearch` does return `E_POINTER` when `that` or `ppv` is null, and documenting it here helps callers who need to guard against null before calling. Also fixes a missing newline at end of file.

GrantMeStrength:
@copilot /review

GrantMeStrength (approved):

PR #2242, opened 2026-07-29
Fix and improve CreateIconFromResource(Ex) and LookupIconIdFromDirectory(Ex) documentation

DJm00n:
CreateIconFromResourceEx:
- presbits: clarify the parameter accepts a single RT_ICON/RT_CURSOR image, not a
  resource group directory; redirect buffer format details to Remarks
- fIcon: remove misplaced LOCALHEADER sentence (covered in the DIB format bullet in Remarks)
- cxDesired/cyDesired: clarify that zero without LR_DEFAULTSIZE uses the size stored in
  presbits without scaling, rather than vaguely "the actual resource size"
- Flags: remove LR_SHARED — it does not produce sharing or caching for this function
  and makes the handle impossible to destroy; remove from keyword metadata as well
- Remarks: replace boilerplate sentence with a proper description of the three accepted
  buffer formats (DIB, PNG, ANI/RIFF-ACON); add ANI sizing and fIcon-ignored notes

CreateIconFromResource:
- presbits: redirect buffer format details to CreateIconFromResourceEx Remarks
- fIcon: remove misplaced LOCALHEADER sentence
- Remarks: remove boilerplate sentence; expand the wrapper description to explain the
  LR_DEFAULTSIZE consequence (always default system size) and how to get a specific size
- Add Examples link

LookupIconIdFromDirectoryEx:
- cxDesired/cyDesired: document that the SM_* metric is only used when LR_DEFAULTSIZE is
  set; without it, zero returns the first directory entry without size scoring
- Flags table: add the missing LR_DEFAULTSIZE entry; fix LR_MONOCHROME description
  ("Searches for" rather than "Creates")
- Add LR_DEFAULTSIZE to keyword metadata

LookupIconIdFromDirectory:
- Description: clarify that this wrapper passes Flags=0 (not LR_DEFAULTSIZE), so it
  always returns the first directory entry without size scoring

Related: https://devblogs.microsoft.com/oldnewthing/?p=108925

### DIFF
--- a/sdk-api-src/content/winuser/nf-winuser-createiconfromresource.md
+++ b/sdk-api-src/content/winuser/nf-winuser-createiconfromresource.md
@@ -65,9 +65,7 @@ To specify a desired height or width, use the [CreateIconFromResourceEx](/window
 
 Type: **PBYTE**
 
-The buffer pointer containing the icon or cursor resource bits. These bits are typically loaded by calls to the [LookupIconIdFromDirectory](/windows/desktop/api/winuser/nf-winuser-lookupiconidfromdirectory), [LookupIconIdFromDirectoryEx](/windows/desktop/api/winuser/nf-winuser-lookupiconidfromdirectoryex), and [LoadResource](/windows/desktop/api/libloaderapi/nf-libloaderapi-loadresource) functions.
-
-See [Cursor and Icon Resources](/windows/win32/menurc/resource-file-formats#cursor-and-icon-resources) for more info on icon and cursor resource format.
+The buffer pointer containing the icon or cursor resource bits. These bits are typically loaded by calls to the [LookupIconIdFromDirectory](/windows/desktop/api/winuser/nf-winuser-lookupiconidfromdirectory), [LookupIconIdFromDirectoryEx](/windows/desktop/api/winuser/nf-winuser-lookupiconidfromdirectoryex), and [LoadResource](/windows/desktop/api/libloaderapi/nf-libloaderapi-loadresource) functions. For the buffer formats this parameter accepts, see the Remarks section of [CreateIconFromResourceEx](/windows/desktop/api/winuser/nf-winuser-createiconfromresourceex).
 
 ### -param dwResSize [in]
 
@@ -80,8 +78,6 @@ The size, in bytes, of the set of bits pointed to by the *presbits* parameter.
 Type: **BOOL**
 
 Indicates whether an icon or a cursor is to be created. If this parameter is **TRUE**, an icon is to be created. If it is **FALSE**, a cursor is to be created.
-
-The [LOCALHEADER](/windows/win32/menurc/localheader) structure defines cursor hotspot and is the first data read from the cursor resource bits.
 
 ### -param dwVer [in]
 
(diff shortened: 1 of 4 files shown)


GrantMeStrength (changes requested):
Technical review: changes requested.

- **Major — factual error (`nf-winuser-createiconfromresourceex.md:119`)**: “other header versions are not accepted” excludes `BITMAPCOREHEADER`, but Raymond Chen’s Microsoft article explicitly documents `BITMAPCOREHEADER` as a legacy format accepted by `CreateIconFromResource(Ex)`: https://devblogs.microsoft.com/oldnewthing/20231025-00/?p=108925. Include the legacy header format or remove the exclusion.

- **Major — needs API-owner verification (`nf-winuser-createiconfromresourceex.md:121`)**: The ANI/RIFF-ACON support, first-size selection, and `fIcon`-ignored behavior are new public contract claims not established by the cited Learn material or the Microsoft icon-format article. Please provide an authoritative implementation/spec source or obtain Windows API-owner confirmation before publishing them.

- **Major — internally inconsistent selection contract (`nf-winuser-lookupiconidfromdirectoryex.md:55` versus `:97`)**: The description says zero dimensions use system metrics, while the proposed `LR_DEFAULTSIZE` row says that without the flag and with both dimensions zero, size does not participate. Qualify the description so system metrics apply only when `LR_DEFAULTSIZE` is set.

- **Major — needs API-owner verification (`nf-winuser-lookupiconidfromdirectoryex.md:55-60`)**: The exact weighted scoring formula and tie-breaking order are reverse-engineered behavioral details. The current public Learn contract describes a different selection model, and the Microsoft article at https://devblogs.microsoft.com/oldnewthing/20121005-00/?p=6393 confirms only that Windows has a default size/color-depth algorithm, not these weights. Please obtain API-owner confirmation or cite an authoritative implementation/spec source.

- **Major — ownership semantics need resolution (`nf-winuser-createiconfromresource.md:98-100`)**: The page says the wrapper passes `LR_SHARED` and then tells callers to destroy the returned handle. `LR_SHARED` normally denotes a cached shared handle that callers must not destroy; the Microsoft resource-semantics explanation is https://devblogs.microsoft.com/oldnewthing/20131003-00/?p=3043. If `CreateIconFromResourceEx` ignores sharing because it has no module/resource identity, state and source that exception explicitly. Otherwise correct the cleanup guidance.

- **Minor — wrong parameter name (`nf-winuser-createiconfromresourceex.md:69`)**: `dwResSize` describes `pbIconBits`, but the function parameter is `presbits`. Replace `pbIconBits` with `presbits`.

DJm00n:
> * createiconfromresource

Minor — wrong parameter name (createiconfromresourceex.md:69). Fixed. pbIconBits → presbits.

Major — factual error, BITMAPCOREHEADER exclusion (createiconfromresourceex.md:119). Fixed. The DIB bullet now says the legacy BITMAPCOREHEADER is also accepted, alongside BITMAPINFOHEADER. Source: Raymond Chen, "The format of icon resources, revisited" (Oct 25, 2023) — https://devblogs.microsoft.com/oldnewthing/20231025-00/?p=108925 — which lists both as valid icon-resource image formats.

Major — internally inconsistent selection contract (lookupiconidfromdirectoryex.md:55 vs :97). Fixed. The description now says system-metric substitution applies only when LR_DEFAULTSIZE is set; when both dimensions are zero without the flag, size does not participate and selection is by color depth alone — matching the LR_DEFAULTSIZE row.

Major — weighted scoring formula (lookupiconidfromdirectoryex.md:55–60). Partially addressed. The formula itself (bpp/width/height deltas, doubled penalty for undersized candidates, color-depth term always doubled, lowest score wins) turns out to be documented by Microsoft: KB 307213, "Windows XP icon and cursor support," Icon selection section — https://web.archive.org/web/20110630025042/http://support.microsoft.com/kb/307213 — spells out exactly this weighting. I've kept the formula and would cite that KB as the authoritative source if needed. I did remove the tie-breaking claim ("greater color depth, then earlier directory entry") — that part isn't in the KB and I don't have an authoritative source for it, so it's gone rather than left unsupported.

Major — ownership semantics, LR_SHARED vs. destroy guidance (createiconfromresource.md:98–100). Fixed, along the line you suggested. Added an explicit explanation: CreateIconFromResourceEx builds the icon/cursor from a memory buffer that has no module or resource identifier, so the shared-resource cache — which is keyed by module + resource ID — never sees this handle even though LR_SHARED is passed internally. The handle is therefore caller-owned, consistent with the existing DestroyIcon/DestroyCursor guidance. Source: Raymond Chen, "The relationship between module resources and resource-derived objects in 32-bit Windows" (Oct 3, 2013) — https://devblogs.microsoft.com/oldnewthing/20131003-00/?p=3043 — which describes the cache as keyed by module and resource ID.

Major — ANI/RIFF-ACON support, first-size selection, fIcon ignored (createiconfromresourceex.md:121). Not addressed — left as-is for now. I don't have a Microsoft Learn/KB source that documents this specific contract, and the Chen icon-resource article above doesn't cover it either (a commenter (me, hehe) on that post flagged the same gap for RT_ANIICON/RT_ANICURSOR). The behavior is reproducible and consistent across the builds I tested, but I understand that's not the bar for a public contract here. Happy to go either way:
- keep it pending an API-owner confirmation you can help get,
- reword it as a "Note" describing observed, undocumented behavior rather than a formal contract, or
- drop the bullet entirely if you'd rather not publish it without sign-off.

Let me know which you'd prefer for that last one, and I'll follow up.

PR #2243, opened 2026-07-30
Fix missing closing italic tag

SimpsonGSD:
Before half of the page was displayed in italics.
Introduced in #2131.


SimpsonGSD:

PR #2253, opened 2026-08-26
GetModuleFileNameExW (return value clarification)

NsPro04:
Based on my tests, if the buffer is too small, the behavior is similar to [GetMappedFileNameW](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getmappedfilenamew#return-value), except that last error is not set to **ERROR_INSUFFICIENT_BUFFER**, and the function returns nSize - 1 instead of nSize.

image
image

### DIFF
--- a/sdk-api-src/content/psapi/nf-psapi-getmodulefilenameexw.md
+++ b/sdk-api-src/content/psapi/nf-psapi-getmodulefilenameexw.md
@@ -89,7 +89,7 @@ The size of the <i>lpFilename</i> buffer, in characters.
 
 ## -returns
 
-If the function succeeds, the return value specifies the length of the string copied to the buffer.
+If the function succeeds, the return value is the length of the string that is copied to the buffer, in characters, not including the terminating null character. If the buffer is too small to hold the module name, the string is truncated to nSize characters (including the terminating null character), the function returns _nSize - 1_.
 
 If the function fails, the return value is zero. To get extended error information, call 
        <a href="/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror">GetLastError</a>.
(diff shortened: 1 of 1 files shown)


GrantMeStrength (changes requested):
Technical review: changes requested.

- **Major — version/platform scope (`sdk-api-src/content/psapi/nf-psapi-getmodulefilenameexw.md:92`)**: The new `nSize - 1` return and unchanged-last-error contract is based on contributor testing, but this API page applies from Windows XP/Server 2003 onward and documents both the Psapi.dll wrapper and the K32 implementation selected by `PSAPI_VERSION`. A test on one current build is not sufficient to promise identical behavior across all supported implementations. Please obtain confirmation from the Windows API owner or provide an authoritative implementation/spec source; otherwise scope the statement to the Windows versions/builds actually tested. The existing comparison with `GetMappedFileNameW` does not establish this function’s contract.

The earlier Copilot concern about explicitly stating that `ERROR_INSUFFICIENT_BUFFER` is not set has been addressed.

NsPro04:
Uhh... ok, I think I can test it on Win XP

NsPro04:
So far, I have tested it on Win 11 - the behavior is similar to Win 10

IMG_20260915_145800_831

(In the Windows 11 console, '\0' isn't displayed, but as we can see from the byte representation, it's actually there.)

NsPro04:
image

Oh, that's funny... but in Windows XP, it returns the correct value. It returns `nSize` as it should.
It was broken later...
Okay, I'll look into it.

NsPro04:
image

In Windows Vista, everything works fine, too. It behaves the same way as Windows XP.
`nSize` is returned.

NsPro04:
image

Windows 7 - same as before

NsPro04:
image

Windows 8.0 - still works fine

NsPro04:
image

Windows 8.1 - same

NsPro04:
image

Windows 10 1507 (if I’m not mistaken, this is the first release)
Starting from this point, the returned value changed to `nSize-1`

NsPro04:
So...
Now the key question is - what should we do with all this? XD
I guess we should just write it as it is:
that before Windows 10, it returned `nSize`, but on Windows 10 and later, it returns `nSize-1`.
Im not sure

PR #2258, opened 2026-09-09
Update Get/SetPriorityClass docs

Zk6Z:
Task List was the name used in Windows 9x. Task Manager is the modern name.


Zk6Z:

Zk6Z:
Seems newer Task Manager in Wimdows 11 uses normal priority.

